PR #45, opened 2014-02-19
Add hooking / code injection category.

AndrejMitrovic:
I guess this would be the first:
http://code.dlang.org/packages/dmadhook

Btw, is there a way to mark a package as being OS-specific?


andralex:
ping?

MartinNowak:
Seems to narrow for now.

PR #76, opened 2014-10-31
Added API framework and a package stats endpoint

ColdenCullen:
The idea is that this should allow us to start adding badges to [shields.io](http://shields.io/).

This PR also leads the way towards adding more stats in the future.


MartinNowak:
thx

MartinNowak:
Shield now available on http://shields.io/.

PR #97, opened 2015-02-19
Search functionality for dub CLI

grogancolin:
Adds a function to api.d to support searching for packges for the dub command line tool.

Used by:
https://github.com/D-Programming-Language/dub/pull/515

It is accessible through:
/api/search?q=<comma,seperated,names>

and returns a json string with format
{
    "searched for string" : [ "closely", "matching", "values"]
}

It uses the m_registry.searchPackages function to do the actual searching, so if/when this is updated to run with elastic search it should slot right in.

### DIFF
--- a/source/dubregistry/api.d
+++ b/source/dubregistry/api.d
@@ -46,4 +46,45 @@ class DubRegistryWebApi {
 		if (stats.type != Json.Type.null_) res.writeJsonBody(stats);
 		else res.writeJsonBody(["message": "Package/Version not found"], HTTPStatus.notFound);
 	}
+
+    // searches for a package
+    void querySearch(HTTPServerRequest req, HTTPServerResponse res){
+        string* q = "q" in req.query;
+        
+        if(q == null) 
+        {
+            res.statusCode = 400;
+            res.writeJsonBody("Error: Must pass query parameter");
+            logDiagnostic("Error: no query in search string");
+            return;
+        }
+
+        string[] queryStrings = (*q).split(",");
+
+        if(queryStrings.length == 0)
+        {
+            res.statusCode = 400;
+            res.writeJsonBody("Error: Must pass search text");
+            logDiagnostic("Error: no query parameters in search string");
+            return;
+        }
+        // limit the number of items can search for, to limit possibility for attacks
+        // should maybe pick a better number here, 15 is arbitrary limit I chose
+        if(queryStrings.length > 15)
+        {
+            res.statusCode = 400;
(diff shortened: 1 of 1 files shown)


MartinNowak:
Nice

source/dubregistry/api.d
```diff
@@ -46,4 +46,45 @@ class DubRegistryWebApi {
 		if (stats.type != Json.Type.null_) res.writeJsonBody(stats);
 		else res.writeJsonBody(["message": "Package/Version not found"], HTTPStatus.notFound);
 	}
+
+    // searches for a package
+    void querySearch(HTTPServerRequest req, HTTPServerResponse res){
```
MartinNowak: Please use tabs and get an editorconfig plugin.

source/dubregistry/api.d
```diff
@@ -46,4 +46,45 @@ class DubRegistryWebApi {
...
+        {
+            res.statusCode = 400;
+            res.writeJsonBody("Error: Must pass query parameter");
+            logDiagnostic("Error: no query in search string");
+            return;
+        }
+
+        string[] queryStrings = (*q).split(",");
+
+        if(queryStrings.length == 0)
```
MartinNowak: You can elegantly handle both errors in one case using [get](http://vibed.org/api/vibe.utils.dictionarylist/DictionaryList.get).
```d
auto queries = req.query.get("q", "").splitter(",");
if (queries.empty)
```
Using [splitter](http://dlang.org/phobos/std_array.html#.splitter) instead of split avoids an allocation.

grogancolin: Yep, thats much nicer. Have merged that in! Thanks.

grogancolin:
I've moved this functionality into the rest service from #103.

The searching functionality is now MUCH simpler.
In dubregistry.api
```
    Json getLikeNames(string name, uint maxResults){
        import std.range : take;
        return m_registry.searchPackages([name])
            .take(maxResults)
            .map!(a => a["name"])
            .array
            .serializeToJson;
    } 
```

I think it's preferable to use the above function over the old querySearch in this pull request?

s-ludwig:
Sounds good. I'd propose a signature, such as `string[] querySimilarPackages(string name, uint max_results)`, which avoids the need for explicit JSON serialization and makes it nicer to use from D code (e.g. using the `RestInterfaceClient`).

grogancolin:
I removed the old fork from my github, so had to re-create it to make the changes. Will close this off.
Continue at #115

PR #101, opened 2015-03-24
update documentation of copyFiles

MartinNowak:
- mention that it copies directories as well
- give an example for matching a dir vs. files in a dir


s-ludwig:
Thanks!

PR #103, opened 2015-03-26
REST interface for API

MartinNowak:
- add /api/packages/:name/latest         to get the latest version of a package
- add /api/packages/:name/info           to get info for a package (same as /packages/:name.json)
- add /api/packages/:name/:version/info  to get info for a package version (:version can be latest)
- add /api/packages/:name/stats          to get stats for a package
- add /api/packages/:name/:version/stats to get stats for a package version (:version can be latest)

### DIFF
--- a/source/dubregistry/api.d
+++ b/source/dubregistry/api.d
@@ -10,17 +10,30 @@ import dubregistry.registry;
 
 import vibe.d;
 
-DubRegistryWebApi registerDubRegistryWebApi(URLRouter router, DubRegistry registry)
+void registerDubRegistryWebApi(URLRouter router, DubRegistry registry)
 {
-	auto settings = new WebInterfaceSettings;
-	settings.urlPrefix = "/api";
-
-	auto webapi = new DubRegistryWebApi(registry);
-	router.registerWebInterface(webapi, settings);
-	return webapi;
+	auto pkgs = new Packages(registry);
+	router.registerRestInterface(pkgs, "/api/packages");
 }
 
-class DubRegistryWebApi {
+interface IPackages {
+	@path(":name/latest")
+	string getLatestVersion(string _name);
+
+	@path(":name/stats")
+	Json getStats(string _name);
+
+	@path(":name/:version/stats")
+	Json getStats(string _name, string _version);
+
+	@path(":name/info")
+	Json getInfo(string _name);
+
+	@path(":name/:version/info")
+	Json getInfo(string _name, string _version);
+}
+
+class Packages : IPackages {
 	private {
 		DubRegistry m_registry;
 	}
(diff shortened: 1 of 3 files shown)


source/dubregistry/api.d
```diff
@@ -10,17 +10,30 @@ import dubregistry.registry;
...
-	router.registerWebInterface(webapi, settings);
-	return webapi;
+	@path(":name/:version/stats")
+	Json getStats(string _name, string _version);
+
+	@path(":name/info")
+	Json getInfo(string _name);
+
+	@path(":name/:version/info")
+	Json getInfo(string _name, string _version);
```
s-ludwig: There is still the shortcut of using `string getLatest(string id)` instead of using an explicit `@path`. I'm not sure if this is the best long-term approach to modeling collections of obejcts, but it works and is short. I have no strong preference here personally.

source/dubregistry/dbcontroller.d
```diff
@@ -132,12 +132,26 @@ class DbController {
 
 	bool hasVersion(string packname, string ver)
 	{
-		auto packbson = Bson(packname);
-		auto verbson = serializeToBson(["$elemMatch": ["version": ver]]);
-		auto ret = m_packages.findOne(["name": packbson, "versions" : verbson], ["_id": true]);
+		auto ret = m_packages.findOne(["name": packname, "versions.version" : ver], ["_id": true]);
```
s-ludwig: I'm trying to remember if there way any subtle reason for the `$elemMatch`, but don't really see any...

s-ludwig:
Should we leave the old "stats.json" routes for backwards compatibility? I'll have a look at the server logs to see if those have been used at all.

s-ludwig:
There are a total of 47 requests to some selected stats.json files. About half of them by Google/Yahoo bots and the rest looks like manual requests from a browser. So I guess no backwards compatibility is needed.

MartinNowak:
The rest probably comes from me, I don't think it's necessary.

MartinNowak:
Once merged, I can send my changes to shields.io.
https://github.com/MartinNowak/shields/commits/dub

s-ludwig:
That would be nice. Would that be using an external service, tough? Should be pretty easy to generate the SVG graphics locally on the registry server. At least we should set up a URL that forwards to shields.io, so that we can switch to an own implementation later, or what do you think?

MartinNowak:
Yes, we could offer some redirects as well.
But placing it on shields is also a small D/Dub advertisement and they already do a good job.

MartinNowak:
Can you please ping me when it's deployed?

s-ludwig:
Done now!

MartinNowak:
Now on http://shields.io.

PR #107, opened 2015-04-10
rename about page to getting started

MartinNowak:
- never knew about that page which
  contained the missing dub introduction


s-ludwig:
Hm, good point. Now that you say it, "About" may actually be something that I would have clicked only as a last resort, too.

MartinNowak:
Someone shared the link in the #d channel, I was pretty surprised to see this, b/c I thought the lack of an introduction tutorial is one reason that hampers dub adoption.

PR #111, opened 2015-04-16
Make table headers clickable for sorting.

s-ludwig:
Also adds a new "Added" column to make this scheme work. This looks a little bulky currently, any ideas for making that prettier (without too much work)?
header-sort

### DIFF
--- a/public/styles/common.css
+++ b/public/styles/common.css
@@ -53,6 +53,8 @@ dd { margin-left: 18pt; margin-top: 3pt; margin-bottom: 6pt; }
 
 .nobreak { white-space: nowrap; }
 
+.dull { color: #888; font-style: italic; }
+
 /********************/
 /* HEADER           */
 /********************/
--- a/public/styles/common.css
+++ b/public/styles/common.css
@@ -200,6 +200,11 @@ h3 {
 	font-weight: bold;
 }
 
+/* sortable table headers */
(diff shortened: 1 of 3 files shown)


PxlBuzzard:
On the topic of dates, it would be great to do some localization / beautification. With a library like [time-elements](https://github.com/github/time-elements) you can let the client choose how to format a date. Dub could use a relative time for the *Updated* column and a local time for *Added*.

s-ludwig:
Oh right, good idea, I always wanted to add relative dates, but completely forgot about them. Maybe the "Current version" and "Updated" columns could also be merged in that process:
<table>
<tr>
<th>Name</th>
<th>Last update</th>
<th>Registered</th>
<th>Description</th>
</tr>
<td>onyx-log</td>
<td>1.0.0-beta.2<em>, updated 2 hours ago</em></td>
<td>2015/04/16</td>
<td>Log library for D</td>
</tr>
</table>

Regarding localized dates, that may look a little out of place as long as the rest of the site isn't localized. Not sure if that makes sense at all, because virtually all content will be in English anyway.

s-ludwig:
Now looks like this:
header-sort2

PxlBuzzard:
:+1: I think that looks good and I like the minimal implementation in D.

s-ludwig:
Okay, I also think that's a decent solution. I'll merge and deploy.

PR #115, opened 2015-07-02
Added search functionality to api.d

grogancolin:
Original discussion in #97

### DIFF
--- a/source/dubregistry/api.d
+++ b/source/dubregistry/api.d
@@ -31,6 +31,9 @@ interface IPackages {
 
 	@path(":name/:version/info")
 	Json getInfo(string _name, string _version);
+
+	@path("search")
+	Json querySimilarPackages(string name, uint maxResults=5);
 }
 
 class Packages : IPackages {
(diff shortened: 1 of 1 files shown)


s-ludwig:
Did you try to use `string[]` as the return type? Should work if you use `.map!(a => a["name"].get!string).array;`.

Regarding the URL, my initial idea was `/similar_packages?q=`, but considering how `searchPackages()` works, that isn't really fitting. I think the `/search?q=` that you have now is the better choice, so the method should probably also be named like that instead (`@method(HTTPMethod.GET) void search(...);`)

grogancolin:
I didn't try string[] as the return type, as I figured in the future it might be worthwhile to return more than just the matching names. It'd be good to have the latest version and description in there too I think, as I can see a user wanting this right after searching so including it would save another request.

One other option I was thinking was to have the columns returned customisable in the request body.
The function would be something like:
```
Json querySimilarPackages(string name, uint maxResults=5, string[] headings = ["name"]);
```

Though, the above probably isn't very RESTful...

I guess this feature has the potential to balloon up, and I should probably just keep it simple :)

grogancolin:
Thinking about it, it's better to keep this simple, can add more functions / REST Resources later if we want to get more information in a search.

Have changed the function to return a string[] now too.

s-ludwig:
Superseded by a more thorough implementation in #121. The visible difference from the outside is that it returns a tuple of name, description and version instead of just the name and that it *doesn't* take a `maxResults` parameter (we could still add that).

Sorry for having left this open for so long, it could have been merged as a first step long ago! I got distracted by other things and then this dropped off the radar.

PR #121, opened 2015-09-20
implement search API

MartinNowak:
- returns {name, description, latest_version} tuples

### DIFF
--- a/source/dubregistry/api.d
+++ b/source/dubregistry/api.d
@@ -16,7 +16,12 @@ void registerDubRegistryWebApi(URLRouter router, DubRegistry registry)
 	router.registerRestInterface(pkgs, "/api/packages");
 }
 
+struct SearchResult { string name, description, version_; }
+
 interface IPackages {
+	@method(HTTPMethod.GET)
+	SearchResult[] search(string q);
+
 	@path(":name/latest")
 	string getLatestVersion(string _name);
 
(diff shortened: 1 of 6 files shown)


source/dubregistry/api.d
```diff
@@ -17,6 +17,8 @@ void registerDubRegistryWebApi(URLRouter router, DubRegistry registry)
 }
 
 interface IPackages {
+	Json querySearch(string q);
```
s-ludwig: I'd prefer a strongly typed (and using `@method`) signature like in #115:
`@method(HTTPMethod.GET) string[] search(string q, uint maxResults=5);` (not sure about the `maxResults` parameter)

Having `Json` in the D API is always prone to silent breakage and field access bugs.

s-ludwig:
Should we return the full package information for each package, or rather just the package name, or maybe something in-between (such as name + description + latest_version)? The full info could waste a lot of bandwidth for common scenarios.

MartinNowak:
I don't think the full info is useful as a search result.

source/dubregistry/api.d
MartinNowak: Will do.

MartinNowak: Why isn't there a `@GET` abbreviation for `@method(HTTPMethod.GET)`?

MartinNowak:
Updated as requested.

s-ludwig:
LGTM. I was thinking about allowing `@method!"GET"` as a shortcut. `@GET` would of course be a lot better still in that regard, but using it is usually rather rare, so it doesn't have to be as concise as possible (at the cost of polluting the global namespace).

PR #126, opened 2016-01-05
Documentation for compiler specific commands in dub.json/dub.sdl

BlackEdder:
This pull request adds information about compiler specific settings in the configuration files. This is especially useful when the preBuildCommands needs to compile/run a d program to generate code/settings before install. 

I used the information provided by Sonke:
http://forum.rejectedsoftware.com/groups/rejectedsoftware.dub/thread/5707/

### DIFF
--- a/views/package_format_json.dt
+++ b/views/package_format_json.dt
@@ -369,12 +369,12 @@ block body
 			td preBuildCommands
 			td
 				code string[]
-			td A list of shell commands that is executed always before the project is built
+			td A list of shell commands that is executed always before the project is built. Can be specific per compiler by appending <code>-&lt;compiler&gt;</code> (i.e. preBuildCommands-dmd, preBuildCommands-ldc, etc.).
 		tr
 			td postBuildCommands
 			td
 				code string[]
-			td A list of shell commands that is executed always after the project is built
+			td A list of shell commands that is executed always after the project is built. Can be specific per compiler similar to <code>preBuildCommands</code>.
 
 		tr
 			td dflags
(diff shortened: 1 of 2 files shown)


s-ludwig:
Since platform specifications work for almost all build settings, I wouldn't want to have them repeatedly mentioned for each setting, but mentioning them only for `*BuildCommands` is also not ideal. Currently they are mentioned at the beginning of the "Build settings" section (JSON) or in a separate sub section "Platform specifications" (SDLang). Maybe there is another way to make them more visible without adding duplication to the docs.

Good that you added the `$DC` variables, I totally forgot about that. There is actually even a bunch of additional ones missing: https://github.com/D-Programming-Language/dub/blob/20d74e88a1b4cf7af001b55d221479b71bd15aaf/source/dub/generators/generator.d#L456

BlackEdder:
I had totally missed the mention of the -dmd etc. under build settings. Maybe this was because it refers to platform, which in my mind means linux/windows/macosx. Mentioning compiler there might be better?

I guess maybe just merge 3a2d81e although ideally all the variables would be mentioned there (or at least a link to the definitions in generator.d would be helpful)

wilzbach:
> I guess maybe just merge 3a2d81e for now? Ideally all the variables would be mentioned there (or at least a link to the definitions in generator.d would be helpful)

ping?

s-ludwig:
3a2d81e is on master now as 92161b6a39d83253adb297271e400984e0db60be. I'll add documentation for the remaining variables, too.

> Maybe this was because it refers to platform, which in my mind means linux/windows/macosx. Mentioning compiler there might be better?

Will have a look at this, too.

s-ludwig:
Documenting all variables: fdcba551c7b92fd00dde58b11dee1f51ff22ee7f
Clarify platform specs: 99da1305af4c3991cb412a41de3a7c8765b0efe7

PR #133, opened 2016-02-11
Mention "unittest" configuration special case

Dicebot:
As explained in IRC by @s-ludwig


s-ludwig:
Eventually there should be a dedicated page about such higher level concerns, but for now this is probably the best place. The text is missing a few articles, I'll fix that post merge (easier that even writing them out here ;)

PR #134, opened 2016-02-12
Mention the new "default" dependency attribute.

s-ludwig:
Accompanying PR for D-Programming-Language/dub#733

### DIFF
--- a/views/package_format_json.dt
+++ b/views/package_format_json.dt
@@ -4,6 +4,8 @@ block title
 	- auto title = "Package file format (JSON)";
 
 block body
+	- enum iff = `<a href="https://en.wikipedia.org/wiki/If_and_only_if" class="blind"><em>iff</em></a>`;
+
 	h2 Introduction
 
 	p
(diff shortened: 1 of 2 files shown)


John-Colvin:
seems reasonably understandable to me

s-ludwig:
OK. Down the line, there should be some kind of high-level documentation/tutorial that explains everything by example etc., but for the format spec, I think this should be sufficient. Maybe it'll need to be adjusted before 0.9.25 gets released, because there should probably be a way to select/deselect optional dependencies form the CLI, rather just by editing dub.selections.json.

I'll merge once Travis finishes.

Abscissa:
Maybe just a nitpick, but I tend to wonder how well-known "iff" is. I'm familiar with it, but I don't know whether it's common-knowledge among programmers. Personally, I tend to avoid it because of that, but maybe I'm just overly cautious.

Also, maybe something should be added to the "default" section to answer the question "What if I use default:true without optional:true, or with an explicit optional:false? Error? Ignores 'default'? Automatically implies optional:true?"

s-ludwig:
I think for anything spec-like, "iff" is a really useful language tool, as the alternative is either blowing up the size of the text, or using imprecise wording. But I agree that it's mostly just known in academia, so I don't know... I'd probably still use it, it's useful to know and easy to look up, but I don't have a strong opinion. 

Good point about the other attribute combinations, I'll add something.

s-ludwig:
It's still somewhat implicative, but I guess it makes it clear enough that `false` could in theory be specified? I didn't want to needlessly blow up the text.

Abscissa:
> I think for anything spec-like, "iff" is a really useful language tool,

I definitely agree. Maybe linking it to 
https://en.wikipedia.org/wiki/Iff would be appropriate? I'd guess people 
unfamiliar with it would be more likely to assume it's a typo than a 
term to look up. Or I could just be overthinking it.

s-ludwig:
OK, links added.

s-ludwig:
BTW, I think you are right, there was actually a typo PR against the vibe.d docs, because of that. This is why I put it inside `<em>`, but a (subtle) link can't hurt.

PR #138, opened 2016-03-15
show dub install to the user

wilzbach:
It would be very nice if we could extend the installation instructions with a command that the users have to run as it is known from most package managers, but most notably `npm`.
I imagine the command to be similar to this:

```
dub fetch <package-name>
```

This also solves [one of the issues with the two different package formats](https://github.com/D-Programming-Language/dub/issues/789).

[Once `dub` supports `--save`](https://github.com/D-Programming-Language/dub/issues/790) it could even be used to replace the `package.json` example.

In case you don't like this suggestion, an alternative would be to have multiple tabs as [Awesome Vim](http://vimawesome.com/plugin/fugitive-vim) does.

### DIFF
--- a/views/view_package.dt
+++ b/views/view_package.dt
@@ -117,6 +117,9 @@ block body
 		|	}
 		|}
 
+	p You can directly install it:
+	pre.code dub fetch #{packageName}
+
 	- if (versionInfo.readme.opt!string.length)
 		h2 Readme
 


greenify:
>  If you mean something like npm install, 

Yes I do - optimally with auto-addition to the local dependency list, but that's a different issue.

> it already has counterpart via dub fetch.

Yes I did - and the PR already mentions `fetch` - just forgot to update the text here ;-)

s-ludwig:
I'd propose to make two extensions:
1. Only suggest the `dub fetch` for packages that have an executable configuration (need to look into how to best implement this)
2. Add `dub run <packagename>` after the fetch so show how the installed package can be used

For libraries, it doesn't make too much sense, at least until we have a fetch-adds-dependency feature (which is a bit difficult to implement nicely without destroying user formatting in dub.(json/sdl)).

wilzbach:
> an executable configuration (need to look into how to best implement this)

Don't we have access to the package's json specification?
So we could just loop over all buildTypes? Or is looking at the first buildType already enough?

> For libraries, it doesn't make too much sense, at least until we have a fetch-adds-dependency feature 

Agreed - that's a separate [issue](https://github.com/dlang/dub/issues/790).

s-ludwig:
>> an executable configuration (need to look into how to best implement this)
>
> Don't we have access to the package's json specification?
> So we could just loop over all buildTypes? Or is looking at the first buildType already enough?

The problematic part is that this also requires knowledge of the folder/file structure, because by default this is inferred rather than listed explicitly in the package recipe. So either the package would have to be downloaded and extracted, or we'd have to add an abstraction layer that allows injecting a virtual directory structure instead of directly accessing the file system and then only query the GitHub API for the directory listing.

MartinNowak:
Isn't this sth. application developers would typically want to add to the first README section.
Not sure this adds enough useability, in particular considering the implementation effort to detect executable packages.

MartinNowak:
Let's see whether this comes up again with the questions solved.

PR #139, opened 2016-03-15
show by default sdl as package format

wilzbach:
As proposed in #789 it makes sense to show the specification for SDLang by default to users.


s-ludwig:
This was actually part of the SDLang->JSON back-transition: 24206efb2e7555cf04b93ab2c84051c773d9a68a

But, I was thinking about merging the two pages and offering a simple JS supported switch to toggle between the formats (maybe remembering the choice in a cookie would also be nice).

wilzbach:
I close this now, because with the new design both package format's are linked in the menu.

PR #141, opened 2016-03-18
use dlang.org's new design

greenify:
Hi all, 

I thought it might be nice to update the registry design within the change to `dub.pm`.
As branding is very important, I used exactly the same css as dlang.org uses. Thanks for the amazing design @CyberShadow! (I hope this is ok for the dlang.org team)

A couple of screenshots:

2016-03-18_22-06-27_scrot

2016-03-18_22-06-48_scrot

2016-03-18_22-08-45_scrot

A couple of points
- dlang.org needs jquery for the menu - this code `menu.js` doesn't use much of jQuery's functionality and using vanilla should be rather easy. However if we plan to add more features anyways, we will probably keep the dependency to jQuery -> therefore imho it makes sense to leave it
- I copied the CSS stylesheet from dlang.org 1:1 without modifications to `top.css`, I made the modifications to `top_p.css`
  - we probably could remove some bits from their stylesheet, however its still very small and my experience shows that touching it will make updates more difficult
  - we should have a nice prebuild tool that concats the tree css tools (what is the best way to do this within the dub pipeline?)
- I only found `dub.xar` as dub's image - however under Linux there seems to be no program that is able to deal with this file format. Hence I remade as an SVG graphic - would be happy to use the "original" graphic if you can convert it and there's a difference ;-)
- the new nested menu generation part could be improved - I just hacked it together (second time I use Diet templates)
- this is not well tested yet
- I really would prefer that we switch this switch __within__ the change to `dub.pm` as main domain (however if that's not foreseeable, let's go now)

Enjoy!


CyberShadow:
> Thanks for the amazing design @CyberShadow!

Wasn't me :)

MartinNowak:
Great, thanks a lot for working on this.
Wasn't there a new dub logo http://code.dlang.org/packages/dub-registry/~new-logo?
I'd like to stay away from jquery if possible. It usually isn't used for much beyond what document.querySelectorAll provides.
This is just a style change and doesn't solve any of the usability issues of the registry, but it shouldn't restrict or ability to overhaul the site, so I'm all for it.

s-ludwig:
Ditto for new logo and jQuery. The new logo is found at the bottom of https://sdlang.org/ (and in b/w at the top of vibed.org). I can of course provide you with higher resolutions, too. The plan is to really publish it with the next DUB version, so synchronizing it with merging this PR would be nice, I think.

I like the design otherwise - except for the serif font. Serif fonts are nice for continuous text, but less ideal for tabular data. They also always have an ancient touch, while I think a more elegant and minimal shape is a better fit for our purposes.

One question about the last screenshot. Can the menu pop up dynamically using that menu button at the top-right? If not, I'd just hide that button until that is implemented.

greenify:
> Semi-OT: I'd prefer code.dlang.org (or some other dlang.org subdomain) to stay as primary domain as it defines clear "official" relation with dlang upstream. dub.pm is a nice shortcut though.

Hmm I see your point. I would prefer `dub.pm`, but that's politics - so I will stop to get involved in that ;-)

> Ditto for new logo and jQuery. 

I removed jQuery as dependency - I needed to "emulate" five methods `hasClass`, `removeClass`, `matches`, `toggleClass`, `parents`. It should still be a lot more lightweight now.

> I can of course provide you with higher resolutions, too.

@s-ludwig do you have the new logo as a SVG graphic?

> The plan is to really publish it with the next DUB version, so synchronizing it with merging this PR would be nice, I think.

Nice idea - do you already have a date for the release?

> I like the design otherwise - except for the serif font. Serif fonts are nice for continuous text, but less ideal for tabular data. They also always have an ancient touch, while I think a more elegant and minimal shape is a better fit for our purposes.

Honestly I like both versions.

Roboto:
2016-03-21_19-01-47_scrot

Lucida Sans:
2016-03-21_19-01-28_scrot

Roboto:
2016-03-21_19-02-19_scrot

Lucida Sans:
2016-03-21_19-02-03_scrot

@ others: You can compare both fonts with Chrome Dev Tools. You can click on the `body` element, browse to the `font-family` setting (`top.css:43`) and untoggle it.

> Can the menu pop up dynamically using that menu button at the top-right? If not, I'd just hide that button until that is implemented.

Yes it behaves exactly as on dlang.org.

I put the code on a cheap VPS at [alpha.dub.pm](http://alpha.dub.pm) - so you can play with it and give me your feedback ;-)

public/scripts/menu.js
```diff
@@ -0,0 +1,86 @@
...
+}
+
+miniDom.matches = function(el, selector) {
+    // shortcut to use matches globally
+    if(typeof selector === "undefined"){
+        selector = el;
+        el = document.body;
+    }
+    return (el.matches || el.matchesSelector || el.msMatchesSelector || el.mozMatchesSelector || el.webkitMatchesSelector || el.oMatchesSelector).call(el, selector);
+};
```
MartinNowak: No need for polyfill. http://caniuse.com/#feat=matchesselector

MartinNowak: Should only need to look for `msMatchesSelector`.
http://caniuse.com/#feat=matchesselector

public/scripts/menu.js
```diff
@@ -0,0 +1,86 @@
+miniDom = {};
+
+miniDom.hasClass = function(el, className){
+    if (el.classList)
+        return el.classList.contains(className);
+    else
+        return new RegExp('(^| )' + className + '( |$)', 'gi').test(el.className);
+}
```
MartinNowak: No need for polyfill.
http://caniuse.com/#feat=classlist

public/scripts/menu.js
```diff
@@ -0,0 +1,86 @@
...
+        elem = elem.parentNode;
+    }
+    return parents;
+};
+
+if (typeof cssmenu_no_js === 'undefined') {
+    var open_main_item = null;
+
+    var expandToggles = document.getElementsByClassName("expand-toggle");
+    [].forEach.call(expandToggles, function(expandToggle){
```
MartinNowak: Weird idiom, would `(expandToggles || []).forEach(function...` work as well?

greenify: `expandToggles` is a HTMLCollection, so it doesn't have the forEach methods ..

public/scripts/menu.js
```diff
@@ -0,0 +1,86 @@
...
+            miniDom.toggleClass(container, 'open');
+
+            /* In the main menu, let only one dropdown be open at a
+            time. Also close any open main menu dropdown when clicking
+            elsewhere. */
+            if (open_main_item !== container && open_main_item !== null) {
+                miniDom.removeClass(open_main_item, "open");
+            }
+            var clicking_main_bar = miniDom.parents(container, "#top").length > 0;
+            var clicking_hamburger = this === miniDom.matches('.hamburger')[0];
```
MartinNowak: Rather than searching for parents here, you could add different callbacks to `#top expand-toggle` and `.hamburger expand-toggle`.

public/scripts/menu.js
```diff
@@ -0,0 +1,86 @@
...
+                open_main_item = miniDom.hasClass(container, 'open')
+                    ? container : null;
+            }
+            return false;
+        });
+    });
+
+    // close window on clicks to other regions
+    window.addEventListener("click", function(e) {
+        var clicking_main_bar = miniDom.parents(e.target, "#top").length > 0;
```
MartinNowak: In fact, just adding the handler to `#top` should catch any event that bubbles up.

MartinNowak: So I think sth. along this line should work.
```js
var menu = document.getElementById('#top');
menu.addEventListener('click', function(e) {
  menu.target.classList.add('open');
  return false;
});
document.getElementsByClassName('expand-toggle').forEach(function (el) {
  el.addEventListener('click', function(e) {
    el.target.classList.add('open');
    return false;
  }
});
window.addEventListener('click', function(e) {
  menu.classList.remove('open');
});
```

MartinNowak: It's a bit oversimplified.
I think we can live w/ the JS for now, and simplify it later if necessary.

public/scripts/menu.js
```diff
@@ -1,57 +1,86 @@
...
+            if (open_main_item !== container && open_main_item !== null) {
+                miniDom.removeClass(open_main_item, "open");
+            }
+            var clicking_main_bar = miniDom.parents(container, "#top").length > 0;
+            var clicking_hamburger = this === miniDom.matches('.hamburger')[0];
+            if (clicking_main_bar && !clicking_hamburger) {
+                open_main_item = miniDom.hasClass(container, 'open')
+                    ? container : null;
+            }
+            return false;
```
MartinNowak: This doesn't seem to work, clicking on the menu button on http://alpha.dub.pm (on mobile) forwards me to http://alpha.dub.pm/menu.html.

views/layout.dt
```diff
@@ -3,56 +3,73 @@ html
...
+		#top
+			.helper
+				.helper.expand-container.active
+					.logo
+						a(href=".")
+							img(id="logo", alt="D Logo", src="/images/dlogo_dub.svg")
+					a(href="menu.html", title="Menu", class="hamburger expand-toggle")
+						span Menu
+					#cssmenu
+						- auto listitems = [tuple("Packages", [""]), tuple("Documentation", ["Getting Started;getting_started", "Development;develop", "Package format (JSON);package-format?lang=json", "Package format (SDLang);package-formati?lang=sdl"]), tuple("About",  ["Forums;http://forum.rejectedsoftware.com/groups/rejectedsoftware.dub", "Bug tracker (website);https://github.com/D-Programming-Language/dub-registry/issues", "Bug tracker (DUB);https://github.com/D-Programming-Language/dub/issues", "Github repository (website);https://github.com/D-Programming-Language/dub-registry", "GitHub repository (DUB);https://github.com/D-Programming-Language/dub"])];
```
MartinNowak: typo `package-formati?lang=sdl`.

public/scripts/menu.js
greenify: should be fixed now

public/scripts/menu.js
greenify: Unfortunately we can't make it that simple as 
1) the code would allow multiple menus to be open at the same time (we need `remove`)
2) the hamburger (menu button for mobile devices is a special case)
3) `return false` is not the entire game - we need to use `stopPropagation` in case we want to prevent it from bubbling up

Thanks to your amazing feedback the menu code now looks a lot cleaner :)

public/scripts/menu.js
greenify: I removed these two lines and just check whether the event target was a `hamburger` - a lot cleaner. However I don't mind to split them up into two different event listeners ;-)

s-ludwig:
This is the new logo in SVG and PNG: [dub-logo.zip](https://github.com/D-Programming-Language/dub-registry/files/183675/dub-logo.zip)
Unfortunately the shadows in the SVG Version didn't get exported correctly, so only the PNG ones are directly usable.

But in any case, I just noticed that the spacing between the page elements got broken with the new style sheet and needs to be tweaked a bit (e.g. the space above a heading is currently smaller than the one below). BTW, did you also try Roboto Sans?

MartinNowak:
> This is the new logo in SVG and PNG: dub-logo.zip

Good timing, I just figured out I can't use .xar files for anything.

MartinNowak:
Made a PR w/ a few fixes https://github.com/greenify/dub-registry/pull/1 and the new logo.

greenify:
> Made a PR w/ a few fixes greenify#1 and the new logo

Thanks a lot. Merged, added two tweaks on top of it:

- new logo on one line for narrow screens
- padding between the logo and menu items

-> updated alpha.dub.pm :)

> just noticed that the spacing between the page elements got broken with the new style sheet and needs to be tweaked a bit (e.g. the space above a heading is currently smaller than the one below)

I did this on purpose - I set the `margin-top` for `h2` elements to `0.5em` instead of `2em`. @MartinNowak already removed that, however imho maybe `1em` or `1.5em` might look nicer?

greenify:
> BTW, did you also try Roboto Sans?

I just realized that a sans-serif font is already used:

```
font-family: "Roboto Slab", sans-serif;
```

The previous font:

```
font-family: 'Lucida Sans Unicode', 'Lucida Grande', sans-serif;
```

So to clarify - what exactly do you dislike about the font? ;-)

s-ludwig:
The "Roboto Slab" family is definitely a serif-style font. At least in large sizes, it looks quite modern, but generally it is a bit heavier that I'd imagine. So even if it's not bad, I'd still like to compare against the normal "Roboto" (and maybe a lighter font weight of the slab variety). BTW, I get a placeholder graphic for the drop-down arrows and the menu button. Seems like a font/unicode subset is missing.

1.5em instead of 2.0em may work for the headings, too, there should just still be a clear visual association of headings with their content.

In the current version, the logo could be a little smaller, so that it's height is about what the D logo is on dlang.org. Its left side should also line up with the page contents. We could also add that planetary arc that is used there to loosen up the appearance a little bit.

greenify:
> I get a placeholder graphic for the drop-down arrows and the menu button. Seems like a font/unicode subset is missing.

Do you experience the same behavior on dlang.org?

s-ludwig:
Nope, works fine there.

greenify:
Roboto Slab 400
---------------------

image

image


Roboto Slab 300
---------------------

image

image


Roboto 400
---------------

image

image


Roboto 300
----------------

image

image


Lucida (previous)
----------------------

image

image

greenify:
> The "Roboto Slab" family is definitely a serif-style font. At least in large sizes, it looks quite modern, but generally it is a bit heavier that I'd imagine. So even if it's not bad, I'd still like to compare against the normal "Roboto" (and maybe a lighter font weight of the slab variety).

I liked Roboto Slab 300 the most.

> BTW, I get a placeholder graphic for the drop-down arrows and the menu button. Seems like a font/unicode subset is missing.

The caret-down/up icons are part of _FontAwesome_  - the font was missing.
I added the font as I expect that we will use more icons and the default unicode carets look pretty ugly. It should be possible to isolate those two icons from FontAwesome as svg and embed them.

> 1.5em instead of 2.0em may work for the headings, too, there should just still be a clear visual association of headings with their content.

done.

> In the current version, the logo could be a little smaller, so that it's height is about what the D logo is on dlang.org. Its left side should also line up with the page contents.

Reduced height to `32px`.

> We could also add that planetary arc that is used there to loosen up the appearance a little bit.

I will try once I find time again.

MartinNowak:
> The caret-down/up icons are part of FontAwesome - the font was missing.
> I added the font as I expect that we will use more icons and the default unicode carets look pretty ugly. It should be possible to isolate those two icons from FontAwesome as svg and embed them.

http://fontello.com/

MartinNowak:
> http://fontello.com/

Done

MartinNowak:
Let's move on w/ this, we can easily improve the logo later.

greenify:
Thanks - looking forward to see this live then :)

MartinNowak:
Can you please take care of updating the site @s-ludwig?

s-ludwig:
I don't think it makes sense to switch the changes live until all issues have been worked out (mostly the navigation bar size/logo placement and the search button). Also it may be a good idea to tie the switch to the next release.

MartinNowak:
What's wrong with the search button? We already added the missing fontawesome icons.
What exactly should we do about the logo?

s-ludwig:
The top bar should  have its contents aligned with the main contents. The logo in particular should start at the same column as the text below. The search button should just have the word "Go" dropped, the loupe is enough and also too close to the "Go" ATM.

Issue #140, opened 2016-03-15
copy+paste for dependency line


wilzbach:
It would be quite nice to have a handy "copy to clipboard" button for the dependency line and hopefully soon to the `dub fetch` (#138).
Other packages manager like [rubygems](https://rubygems.org/gems/jekyll) or even github themselves provide this handy feature.

There is [clipboardjs](https://clipboardjs.com/), so adding this is easy and I would be happy to send a PR. The only question is how do you want me to integrate the bower dependency?
Are there some hooks in dub with which we could run `bower` (or `npm`)?

Related to #96 (copy+paste for shields).

MartinNowak:
I wouldn't want a frontend dependency manager for such a simple addition, just copy the script.
We might choose to use bower for the site makeover.

greenify:
> I wouldn't want a frontend dependency manager for such a simple addition, just copy the script.

Fair enough 

> We might choose to use bower for the site makeover.

Yep my question was in some sense more general - how does one integrate bower or npm with `dub`?
See #792

PR #142, opened 2016-03-21
add copy-to-clipboard buttons

### LINKED ISSUE #96: provide a copy+paste link to add a dub shield to a project
A few projects start to add backlink shields to code.dlang.org.
We should provide a common method to create such shields, that shows the latest version.

https://github.com/kiith-sa/dmarkdown
https://github.com/MartinNowak/qcheck

A nice extra feature would be to copy the `"qcheck": "~>0.1.0"` dependency line into the clipboard when clicking the link.

greenify:
adds [clipboard.js](https://clipboardjs.com/) to allow copy & paste for a package's dependency and shields.

### Package dependencies:

2016-03-21_22-42-24_scrot

### and for shields:

2016-03-21_22-43-18_scrot

update: fixes #96 and #140


s-ludwig:
I would have added the shields to `my_packages.package.dt`, because I think that they are mostly useful for the authors themselves to link to their package. It might also make sense to put actual HTML `<a></a>` into the text box.

The clipboard additions in general and the dependency line change looks good! Having dub.sdl handled there was definitely overdue.

views/view_package.dt
```diff
@@ -137,3 +144,24 @@ block body
 		- else
 			a(href='#{req.rootDir}packages/#{urlEncode(packageName)}/#{vs}')= vs
 
+	h2 Shields
+
+	#copy-shields
+		div
+			.img-wrapper
+				img(src="https://img.shields.io/dub/v/vibe-d.svg", alt="Dub version")
```
s-ludwig: This should probably use `/#{packageName}.svg"`. Same for the one below.

greenify:
> I would have added the shields to my_packages.package.dt, ...

Excellent point!

> It might also make sense to put actual HTML <a></a> into the text box.

I am not so sure on this point as shields are mostly used in the README.md and there markdown is used. Moreover bundling it into HTML tags requires mor chars and thus scrolling. 
On the other hand `<a>` tags work fine in markdown too, we already can fill out `alt` tags and for the majority of users it is more convenient.
(I changed it now)

greenify:
Forgot to remove the shields from `view_package` when I moved them to `my_packages.package` - changed ;-)

s-ludwig:
Okay, thanks, looks good to go!

PR #144, opened 2016-04-04
Redesign part 2

greenify:
(follow-up to #141)

> The top bar should have its contents aligned with the main contents. The logo in particular should start at the same column as the text below. 

Done.

> The search button should just have the word "Go" dropped, the loupe is enough and also too close to the "Go" ATM.

I didn't update the alpha server after @MartinNowak PR - it was just "Go". This PR replaces the word "Go" back to the loupe.

@s-ludwig Anything else that's blocking the new design?

(Changes are uploaded to alpha.dub.pm again - we might want to create some CI at some point)

### DIFF
--- a/public/styles/top_p.css
+++ b/public/styles/top_p.css
@@ -54,3 +54,17 @@ h2 {
   font-weight: normal;
   font-style: normal;
 }
+
+/* minimal font awesome CSS */
+
+.fa {
+  display: inline-block;
+  font: normal normal normal 14px/1 FontAwesome;
+  font-size: inherit;
+  text-rendering: auto;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+}
+.fa-search:before {
+  content: "\f002";
+}
--- a/public/styles/top_p.css
+++ b/public/styles/top_p.css
@@ -16,8 +16,19 @@ table {
 	float: none;
 	margin-right: 0;
 	display: inline-block;
-	margin-left: -15px;
+	margin-left: 0px;
 	padding-right: 15px;
+}
+
+/* let dlang header have same padding as content */
+
+@media (min-width: 600pt) {
+    #top > .helper {
+        padding-left: 10%;
+        padding-right: 10%;
+        margin: 0;
+        max-width: initial;
+    }
 }
 
 @media only screen and (max-width:54em) {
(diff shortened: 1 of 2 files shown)


s-ludwig:
Thanks! Looks good to me now. Just one thing I noticed is that the download menu entry appears to be missing. I'd simply replace the "Register" one with "Download", because the user gets the option to register after clicking "Log in" as well.

greenify:
> I'd simply replace the "Register" one with "Download", because the user gets the option to register after clicking "Log in" as well.

Done.

s-ludwig:
Okay, I think we are good to go. Any opinions on the announcement process? As said I'd like to make it part of a bigger announcement for 0.9.25 (last release before locking down the API in 1.0.0).

BTW, https://github.com/D-Programming-Language/dub/pull/785 should really get in 0.9.25, too. The diff is a lot to read, but contains almost only trivial changes, so if anyone has some spare time, it would be great to get some basic review for that.

greenify:
> Okay, I think we are good to go. Any opinions on the announcement process? 

Dlang announce forum and tweeting with Dlang's twitter account? Maybe reddit?

wilzbach:
> Okay, I think we are good to go. Any opinions on the announcement process?

Maybe posting a link to Slashdot and hackernews?

s-ludwig:
Dunno, maybe that would be more something for the 1.0.0 release? At least my gut feeling is that this might be not relevant enough for the casual observer to gain enough traction. We could of course also still hold off and combine everything into one big announcement.

wilzbach:
>  Dunno, maybe that would be more something for the 1.0.0 release? 

Oh overread the 0.9.25 part - yep big splash 💦 for 1.0!

> We could of course also still hold off and combine everything into one big announcement.

Hmm better release early and often :)

s-ludwig:
So big splash or early release then? :P

So just to recap, 1.0.0 is supposed to come very quickly after 0.9.25 and should ideally only be a bugfix release. Maybe @MartinNowak still manages to implement the new dependency resolver, but since that is not a breaking change, we can also do that for 1.1.0 instead. We could also directly go for 1.0.0, but then we'd have to do some extra marketing to get people to test the betas. Would be really bad to have obvious issues in the big 1.0.0 release.

wilzbach:
> So big splash or early release then? :P

Small splash for now (d forum?) and big splash for 1.0?

PR #148, opened 2016-05-14
fix menu wrap on narrow screens

MartinNowak:
- for screens above 66em (where we fold the search box) and below
  ~1100px the menu entries wrap to another line
- somewhat normal format for a 24" vertical screen
- fixed by reducing the extra wide search field

before
menu
after
menu


wilzbach:
LGTM - Nice fix!

PR #154, opened 2016-05-24
Fixed logo href, not working in nested pages

bausshf:
When clicking the logo in nested packages such as viewing a package, clicking the logo would result in a 404, because it wouldn't redirect to the root.

### DIFF
--- a/views/layout.dt
+++ b/views/layout.dt
@@ -16,7 +16,7 @@ html
 			.helper
 				.helper.expand-container.active
 					.logo
-						a(href=".")
+						a(href="..")
 							img(id="logo", alt="DUB Logo", src="/images/dub-header.png")
 					a(href="#", title="Menu", class="hamburger expand-toggle")
 						span Menu


views/layout.dt
```diff
@@ -16,7 +16,7 @@ html
 			.helper
 				.helper.expand-container.active
 					.logo
-						a(href=".")
+						a(href="..")
```
wilzbach: How about `/`?

s-ludwig: Ideally, `#{req.rootDir}`, which gives the relative path to root.

bausshf: `/` is shorter though, but `#{req.rootDir}` seems more standard. Which would be preferred? I assume the latter?

s-ludwig: It wouldn't matter for code.dlang.org, but since this might be run in an intranet setting and possibly put under a subfolder using a reverse proxy setting, rather `#{req.rootDir}` should be used.

wilzbach:
It seems like people keep bumping into this (see e.g. #157), so let's try to get this merge rather soon.
@bausshf do you mind to change your code and rebase?

PR #156, opened 2016-05-24
Use normal font-weight.

epi:
This makes the site look consistent with main dlang.org site
and also much more readable.

### DIFF
--- a/public/styles/top_p.css
+++ b/public/styles/top_p.css
@@ -41,11 +41,6 @@ table {
 }
 
 
-/* thinner font */
-body, .d_decl .quickindex {
-    font-weight: 300;
-}
-
 h2 {
     margin-top: 1.5em;
 }


public/styles/top.css
```diff
@@ -38,7 +38,7 @@ border-radius: always 4px
 
 */
 
-@import url(https://fonts.googleapis.com/css?family=Roboto+Slab:300,400,700);
+@import url(https://fonts.googleapis.com/css?family=Roboto+Slab:400,700);
```
wilzbach: Our idea actually was that we don't modify the top.css, so that it is easier to keep in sync with the one from dlang.org

public/styles/top.css
```diff
@@ -1272,7 +1272,6 @@ body#Home #content > .intro > div > div > *
 body#Home #content > .intro .pitch
 {
     font-size: large;
-    font-weight: 300;
     margin: auto;
```
wilzbach: As said this is copied from dlang.org, so it actually should be consistent.

s-ludwig:
Instead of from "top.css", this has to be removed from "top_p.css".

I'd still like to improve the font selection over main dlang.org. But since many seem to have reading issues, I agree that we should make this change as a first step.

epi:
Yes, only removing that part from top_p.css affected the look of the site. I modified top.css as well just for completeness.

s-ludwig:
Oh, didn't notice that you also changed top_p.css. So in that case:
> this has to be removed **just** from "top_p.css"

;-) As Sebastian said, the idea is to keep "top.css" identical to dlang.org's version as a baseline, so that we can easily follow style changes from there.

epi:
Ok, now the change is only in top_p.css

s-ludwig:
Okay, thanks!

PR #166, opened 2016-06-23
Remove registered date from home view

wilzbach:
Imho the registered date isn't very important when looking at the package overview and increases the "visual clutter".

Probably the download counts would be, but that would require more queries?
Is there any other good replacement that is a good indicator for a package?

### DIFF
--- a/views/home.dt
+++ b/views/home.dt
@@ -36,7 +36,7 @@ block body
 
 	table
 
-		- static immutable browse_kinds = ["name;Name", "updated;Last update", "added;Registered", ";Description"];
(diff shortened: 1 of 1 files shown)


s-ludwig:
It can be very useful for looking at the list of recently registered packages. Maybe we can move it to the right instead?

Yeah, for download statistics we'd first have to duplicate the numbers within the "packages" table to not affect performance. But there are other potentially useful metrics (e.g. popularity or rating). Maybe we need a UI change and have a drop-down or similar.

andralex:
will close this for now

PR #173, opened 2016-08-25
align package description text to the left

BBasile:
_justify_ looks bad for several packages:

justbad


s-ludwig:
Agree. Doesn't really make sense inside usual table cells.

PR #186, opened 2016-12-27
Add release-debug, release-nobounds and profile-gc build types to documentation.

p0nce:
Add release-debug, release-nobounds and profile-gc build types to documentation.

### DIFF
--- a/views/package_format_json.dt
+++ b/views/package_format_json.dt
@@ -578,6 +578,10 @@ block body
 			td Emit profiling code
 			td -profile
 		tr
+			td "profileGC"
+			td Emit GC profiling information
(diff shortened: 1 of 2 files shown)


s-ludwig:
Thanks, I never thought of these when I added new build modes! This would really be a nice candidate for procedural generation. I'll refactor DUB to have the list of standard built modes exposed somehow.

p0nce:
Yes, build types are in approx. 5 different places (workspace-d, dub command-line help, dub sublime generator, dub registry, package_.d)

It is tempting to automate but if dub is poised to stabilize, having duplicated doc may not be a huge problem.

PR #187, opened 2017-02-24
Implement a simple registry mirror mode

s-ludwig:
Can be enabled with `./dub-registry --mirror https://code.dlang.org/`

The most obvious improvement will be to add an API endpoint to receive a complete database dump instead of querying each package individually.

### DIFF
--- a/source/app.d
+++ b/source/app.d
@@ -29,6 +29,7 @@ void startMonitoring()
 {
 	void monitorNewVersions()
 	{
+		sleep(2.minutes());
 		while(true){
 			if (s_mirror.length) s_registry.mirrorRegistry(URL(s_mirror));
 			else s_registry.checkForNewVersions();
--- a/source/app.d
+++ b/source/app.d
@@ -64,6 +64,7 @@ shared static this()
 	BitbucketRepository.register();
 
 	auto router = new URLRouter;
+	if (s_mirror.length) router.any("*", (req, res) { req.params["mirror"] = s_mirror; });
 	router.get("*", (req, res) { if (!s_checkTask.running) startMonitoring(); });
 
 	// VPM registry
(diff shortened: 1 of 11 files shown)


s-ludwig:
Opened a separate PR to add a more efficient API: #188

source/app.d
```diff
@@ -22,13 +23,16 @@ import vibe.d;
 Task s_checkTask;
 DubRegistry s_registry;
 DubRegistryWebFrontend s_web;
+string s_mirror;
 
 void startMonitoring()
 {
 	void monitorNewVersions()
 	{
+		sleep(2.minutes());
```
s-ludwig: I mostly added this as a throttling mechanism in case the process gets restarted repeatedly (maybe in case of a crash or manually during development). But you are right that this may not always be desired. Another reason is that adding code for cache warmup at application start is planned and that together with the update process would produce quite high load.

I don't know, maybe this can just be made configurable.

source/dubregistry/mirror.d
```diff
@@ -0,0 +1,82 @@
...
+import std.encoding : sanitize;
+import std.format : format;
+
+
+void mirrorRegistry(DubRegistry registry, URL url)
+nothrow {
+	logInfo("Polling '%s' for updates...", url);
+	try {
+		bool[string] current_packs;
+		auto packs = requestHTTP(url ~ Path("packages/index.json")).readJson();
```
s-ludwig: Will add.

s-ludwig:
@wilzbach Okay, if you don't mind, I would merge this now and then add the necessary adjustments in #188. We can then also find a final decision for the startup behavior.

wilzbach:
> @wilzbach Okay, if you don't mind, I would merge this now and then add the necessary adjustments in #188. We can then also find a final decision for the startup behavior.

I don't mind at all. I was just trying to provide some feedback ;-)
Please go ahead!

PR #188, opened 2017-02-24
Add an /api/packages/dump REST API endpoint.

s-ludwig:
This fetches the whole package database in a single request. The response is forced to be GZIP compressed to avoid excessive bandwidth use (currently 20MB uncompressed for ~1000 packages, 1.3MB compressed).

### DIFF
--- a/source/dubregistry/api.d
+++ b/source/dubregistry/api.d
@@ -24,12 +24,31 @@ void registerDubRegistryAPI(URLRouter router, DubRegistry registry)
 {
 	auto pkgs = new LocalDubRegistryAPI(registry);
 	router.registerRestInterface(pkgs, "/api");
+	router.get("/api/packages/dump", delegate(req, res) => dumpPackages(req, res, registry));
 }
 
 /// Compatibility alias.
 deprecated("Use registerDubRegistryAPI instead.")
 alias registerDubRegistryWebApi = registerDubRegistryAPI;
 
+
+private void dumpPackages(HTTPServerRequest req, HTTPServerResponse res, DubRegistry registry)
+{
+	import vibe.data.json : serializeToPrettyJson;
+	import vibe.stream.wrapper : StreamOutputRange;
+
+	res.contentType = "application/json; charset=UTF-8";
+	res.headers["Content-Encoding"] = "gzip"; // force GZIP compressed response
+	auto dst = StreamOutputRange(res.bodyWriter);
+	dst.put('[');
+	bool first = true;
+	foreach (p; registry.getPackageDump()) {
+		if (!first) dst.put(',');
+		else first = false;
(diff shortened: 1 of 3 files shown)


source/dubregistry/api.d
```diff
@@ -24,13 +24,32 @@ void registerDubRegistryAPI(URLRouter router, DubRegistry registry)
...
+
+	res.contentType = "application/json; charset=UTF-8";
+	res.headers["Content-Encoding"] = "gzip"; // force GZIP compressed response
+	auto dst = StreamOutputRange(res.bodyWriter);
+	dst.put('[');
+	bool first = true;
+	foreach (p; registry.getPackageDump()) {
+		if (!first) dst.put(',');
+		else first = false;
+		serializeToPrettyJson(&dst, p);
```
wilzbach: 1) Is pretty printing really appropriate here?
2) Why don't you allow the output to be serialized to Bson?

s-ludwig: I checked and the overhead in the compressed output is negligible (the formatting gets compressed perfectly because it is exactly the same for every entry). On the other hand this makes working with the dump a lot more practical (you'd have to find a JSON formatter or viewer that can cope with large files first).

Using BSON would mostly just complicate the client requirements and is probably not more compact in the end (the biggest advantage of BSON is being able to store compact binary data and that jumping over whole values is O(1)). But in a future revision, letting the client control this with an "Accept" header would be an option.

wilzbach: > But in a future revision, letting the client control this with an "Accept" header would be an option.

Yeah I was just curious ;-)
If this becomes a bottleneck, there are a lot of other strategies (incremental, caching, ...) that we could try as well which could potentially save more time / resources, so l am not advocating to do "premature optimization" here.

s-ludwig: True, using some kind of diff mechanism would be good. I'll open a ticket to remember,

s-ludwig:
Updated mirror mode: #190

MartinNowak:
What's the goal of this?
Definitely useful for testing, but maybe not the best fit for a fallback replication (in comparison to db level replication).

MartinNowak:
In fact when trying to improve site availability we should definitely reconsider PaaS hosting w/ builtin load-balancing, local data centers, and replicated/distributed dbs at hand.
http://alternativeto.net/software/heroku/

wilzbach:
> What's the goal of this?

That someone in e.g. china can easily setup a local DUB mirror.

> Definitely useful for testing, but maybe not the best fit for a fallback replication

Well it doesn't dump the user data - so it's absolutely no replacement for this ;-)

MartinNowak:
Maybe let's align current ideas/goals, e.g. in here https://dlang.slack.com/messages/dub/.

PR #193, opened 2017-03-03
index categories and readme for search

MartinNowak:
- more/better search results, but also more noise

### DIFF
--- a/source/dubregistry/registry.d
+++ b/source/dubregistry/registry.d
@@ -236,7 +236,7 @@ class DubRegistry {
 		auto nfo = v.info.get!(Json[string]).dup;
 		nfo["version"] = v.version_;
 		nfo["date"] = v.date.toISOExtString();
-		nfo["readmeFile"] = v.readme;
+		nfo["readme"] = v.readme;
 		nfo["commitID"] = v.commitID;
 
 		PackageVersionInfo ret;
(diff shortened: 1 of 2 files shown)


MartinNowak:
There isn't much wrong about MongoDB's FTS, it eliminates stopwords, uses a stemmer, weights different fields, and more advanced queries like [Xapian's Query Parser](http://getting-started-with-xapian.readthedocs.io/en/latest/concepts/search/queryparser.html) don't really seem necessary to search for packages.
So let's see how far we get by also indexing the readme and categories.
This will amplify the problem that some quality/rating/downloads metric isn't yet part of the sort order, so it's a bit trickier to deal with not so good search matches.

Unless the urlcache for those README.mds is full, it will take quite a while until the internal package cache is populated and the index page (but only that one) won't render until all packages are cached. But the urlcache is populated, so warming up the cache should only take a few seconds.

MartinNowak:
Any chance to look at this @s-ludwig?

MartinNowak:
I'd move on with this to improve search, now that the ranking feature is merged. Will the latter be deployed soon @s-ludwig?

s-ludwig:
Do you mean this: http://code.dlang.org/?showrating ? This has been deployed right after the merge.

source/dubregistry/registry.d
```diff
@@ -399,6 +393,10 @@ class DubRegistry {
 			return false;
 		}
 
+		if ("description" !in info.info || "license" !in info.info) {
```
MartinNowak: That reordering is to fix missing READMEs on code.dlang.org for packages with errors. It seems that most people simply keep ignoring errors anyhow.

MartinNowak:
Updated, it's just about 80MiB (8MiB compressed) for all READMEs, so that's really not much of a problem (hardly even for the inefficient mirroring).

source/dubregistry/registry.d
```diff
@@ -250,22 +250,23 @@ class DubRegistry {
...
-				ret = data.readAllUTF8();
-			});
-			logDebug("reading readme file done");
-		} catch (Exception e) {
-			logDiagnostic("Failed to read README file (%s) for %s %s: %s",
-				file, version_info["name"].get!string,
-				version_info["version"].get!string, e.msg);
+		auto readme = version_info["readme"].opt!string;
+
+		// compat migration, read file from repo if README hasn't yet been stored in the db
```
wilzbach: shouldn't we store it in the DB too then?

source/dubregistry/registry.d
wilzbach: Yeah it would have been good to check this when/before the package is registered ...

source/dubregistry/registry.d
MartinNowak: To much work to change all the places, it just takes an hour or so to populate all packages in the regular update loop.

MartinNowak:
orm
grafik
vs.
grafik

source/dubregistry/registry.d
MartinNowak: And possibly even earlier, e.g. maybe by integrating a lint function into dub.

PR #196, opened 2017-04-01
Rewrite Getting Started guide.

rjframe:
This is a complete rewrite of the getting started guide (which has become a full "How to use DUB" reference and included items not important to someone just getting started). An "Advanced Usage" (or something similar) page should likely be created as well; if desired I could make that part of this PR, or could make another.

I completely removed:
* "Building foreign projects" - the documentation is incorrect (dub fetch isn't supported) and not particularly useful here.
* The suggested project layout. I do show the generated minimal layout.
* The single-file package documentation. This would be a better fit for an advanced usage section. I do not see this documented anywhere else.
* The local package documentation. This would also be a better fit for an advanced usage section. This is documented in docs/commandline in a reference format.

### DIFF
--- a/views/getting_started.dt
+++ b/views/getting_started.dt
@@ -1,116 +1,119 @@
 extends layout
 
 block title
-	- auto title = "Using DUB";
+	- auto title = "Getting Started With DUB";
 
 block body
-	h2 Introduction
+	h2 Installing DUB
 
-	p DUB is a build tool for D projects with support for automatically retrieving dependencies and integrating them in the build process. The design emphasis is on maximum simplicity for simple projects, while providing the opportunity to customize things when needed.
+	p DUB is the D language's official package manager, providing simple and configurable cross-platform builds. DUB can also generate VisualD and Mono-D package files for easy IDE support.
 
-	p An important property of the build description for each project is that it only contains the minimally needed information to describe how a package is built, skipping things such as build type (debug, release, unittest etc.). What is specified can be done so in a mostly platform and compiler agnostic way using abstract fields (e.g. import paths or used libraries) and by having DUB automatically translate common DMD command line flags to GCC or LDC flags.
+	p To install DUB, search your operating system's package manager or <a href="download">download</a> the pre-compiled package for your platform. The Windows installer will perform all installation steps; for other archives, you will want to ensure the DUB executable is in your path. Installation from source on other platforms is as simple as installing the dmd development files and your system's libcurl-dev, then running <code>./build.sh</code> in the repository's folder.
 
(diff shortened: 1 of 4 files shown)


views/getting_started.dt
```diff
@@ -1,116 +1,109 @@
...
-
-	p To get the package, either use <code>git clone</code> with the projects git URL, or type <code>dub fetch &lt;package name&gt;</code> to have the package downloaded and installed in a package repository in your user folder. <code>dub run &lt;package name&gt;</code> can then be used to build and execute the package. <code>dub fetch --local &lt;package name&gt;</code> can be used instead to extract the package into a sub folder of the current working directory.
-
-	h2#own-projects Creating an own project
-
-	p The easiest way is to type <code>dub init &lt;package name&gt;</code> on the command line to have DUB create an application skeleton for you. It will create the following directory structure:
+	p Notice that there are two configuration file formats available. JSON is a commonly-known format, and SDL (SDLang) is a clean, minimalist format. Both offer equivalent functionality (though unlike JSON, SDLang allows comments); use whichever you prefer.
```
wilzbach: Nit: I would use a dash here

s-ludwig:
I mainly still see these issues:

> I completely removed:
>
> - "Building foreign projects" - the documentation is incorrect (dub fetch isn't supported) and not particularly useful here.

"dub fetch" works fine for me and I think the contents are basically still valid. Did you maybe hit a bug there?

> - The suggested project layout. I do show the generated minimal layout.

No objections.

> - The single-file package documentation. This would be a better fit for an advanced usage section. I do not see this documented anywhere else.

This is actually something that has been very well received and it's often used when showing examples, which makes it a pretty fundamental feature. However, I'd agree that it shouldn't be placed too prominently, so that the full package format is communicated clearly as the default choice.

> - The local package documentation. This would also be a better fit for an advanced usage section. This is documented in docs/commandline in a reference format.

Agree with the advanced section, but let's create an advanced page and move those things there now instead of just removing it. It's unlikely that it will ever be resurrected otherwise.

views/getting_started.dt
rjframe: Instead of the semicolon (preceding "use whichever you prefer"? If not, where would you put it?

rjframe:
@s-ludwig `dub fetch` is working for me now; I've added it back (not pushed yet).

I'll add building foreign projects to the end - the outline will be 1) creating a package, 2) publishing a package, 3) building a published package.

Advanced section - I'm not sure about a name; "advanced usage" sounds like command-line documentation, which it won't wholly be. "Doing More with DUB" sounds catchy, but as a reader, I'm not sure it would be obvious what I'd expect to find. Thoughts/preferences?

PR #197, opened 2017-04-25
Update categories.json

Nafees10:
Added category for "User interfaces" and "Text user interfaces".  
  
I need "text user interfaces" for my package qui

### DIFF
--- a/categories.json
+++ b/categories.json
@@ -48,7 +48,9 @@
 		{"name": "filesystem", "description": "File system libraries"},
 		{"name": "gamedev", "description": "Libraries aimed at game development"},
 		{"name": "graphics", "description": "2D/3D graphics"},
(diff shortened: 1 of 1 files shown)


WebFreak001:
You need to change the name then too (call them ui and tui) so that people can add them. It might even be better to create ui and add gui and tui as children but I'm not sure how well that is gonna work with backwards compatibility @s-ludwig ?

MartinNowak:
> It might even be better to create ui and add gui and tui as children but I'm not sure how well that is gonna work with backwards compatibility @s-ludwig ?

Not without a migration, hierarchical categories suck.

MartinNowak:
Merged in fe18df23b32f602a7579816a801a8a4e98346b5c

Issue #169, opened 2016-07-02
Usage statistics for libraries. (Stars + Downloads?)


schancel:
Lots of package repositories (e.g. rubygems) show the number of downloads of a particular package, and there's some way to rate a library.   This would be extremely helpful for sorting through the libraries.  Also, the categories are a little weak.  It would be great to have some way to tag libraries/categorize libraries by their use.

If these things of statistics are amenable to the current project owners here, I'd be interested in working on them.

MartinNowak:
I already have a few plans to restructure categories and searching.
We do have the download statistics in place (http://code.dlang.org/api/packages/vibe-d/stats), but they need to be integrated in the search results.
There is some WIP code for moving all the package sorting into the db, https://github.com/MartinNowak/dub-registry/commit/41a74a216be6b8aa9d26b0e644f2d8536015e718, number of downloads could just be another sortable column.
We thought about some more complex rankings, but I think download numbers are a really good proxy for any other metric.

9il:
Also the list of packages that is use for example Mir would be very helpful. So people would be able to see if their packages are used by another library. see #172

MartinNowak:
Fixed by #224, https://code.dlang.org/?showrating

PR #201, opened 2017-05-17
Display download statistics on 'Edit package' page

ohdatboi:
Looks like this:
dub registry stats mypackage

It works only on 'Edit package' page which means only package owners can see it. I don't want to add this to main package page `'view_package.dt'` yet, because it looks ugly.
Implements #169 (at least partially).


s-ludwig:
While the goal should definitely be to have this publicly visible for each package in a nice form (a graph would be even nicer than the day/week/month accumulations), I don't see anything that would speak against adding it here temporarily in simpler form!

PR #206, opened 2017-05-27
Use pagination on the full package list

s-ludwig:
This is an intermediate step to improve loading performance and bandwidth use of the registry home page. This will later be replaced by different top ten lists.

### DIFF
--- a/public/styles/common.css
+++ b/public/styles/common.css
@@ -158,6 +158,22 @@ h3 {
 	border: 1px solid #ddd;
 }*/
 
+#content ul.pageNav {
+	list-style: none;
+	padding: 0;
+	margin: 0;
+}
+ul.pageNav li {
+	display: inline-block;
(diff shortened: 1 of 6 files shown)


wilzbach:
I think we seriously need to improve the stylesheet of the pagination list:

image

s-ludwig:
Outdated CSS in cache? This is how it looks for me:
image

wilzbach:
> Outdated CSS in cache? This is how it looks for me:

Oh sorry - yeah it might be good to avoid this for other users as well by generating a hash-summed CSS file during a build.

wilzbach:
> it might be good to avoid this for other users as well by generating a hash-summed CSS file during a build.

I opened an issue, s.t. it doesn't get forgotten:

https://github.com/dlang/dub-registry/issues/210

PR #208, opened 2017-05-28
Let getPackageInfo return a static type and fix access to "url".

s-ludwig:
This makes another step into the direction of working only with static types. In particular, this also fixes the issue of accessing the non-existent JSON field "url" (removed after deprecation period), which didn't get caught because it was read from a dynamic 'Json' value.


s-ludwig:
For reference, this is the "url" field that got removed: https://github.com/dlang/dub-registry/commit/5e8c45bdea98a717854e47f8c704e9f53fedb73a#diff-31fe65cc39b31f74454f75b09e88a761L223

PR #211, opened 2017-05-28
Fix potential server crash

WebFreak001:
Server (code.dlang.org) crashes if you open [/?limit=0](https://code.dlang.org/?limit=0)


source/dubregistry/web.d
```diff
@@ -60,6 +60,9 @@ class DubRegistryWebFrontend {
 		import std.algorithm.comparison : min;
 		import std.algorithm.iteration : filter, map;
 
+		if (limit == 0) // prevent division by zero
```
wilzbach: `limit < 0` ?

WebFreak001: it's a ulong

Issue #210, opened 2017-05-28
Introduce a build process for resources


wilzbach:
- [ ] Introduce a build process executable and add it to `postBuildCommands`
- [ ] Bundle and uglify JavaScript
- [ ] Bundle and minimize CSS
- [ ] Save both bundles with a unique hash as part of the file name (can even be the git hash) and pass the bundled file names to `layout.dt` (could be by simply using the newest files in a specific folder, writing the files into a special file name, ...)

s-ludwig:
At least with the new CTFE engine it would also be nice to do this at compile time.

PR #215, opened 2017-06-01
New layout of 'Add package' page

ohdatboi:
This PR changes layout of the 'Add package' page. I wanted to make it mobile-friendly and clear.

### Screenshots:
big
big_error
small

### DIFF
--- a/public/styles/common.css
+++ b/public/styles/common.css
@@ -344,4 +344,47 @@ ul.unmarkedList > li:hover,
 	padding: 0.1em 0.3em;
 }
 
-span.light { opacity: 0.8; }
+.light { opacity: 0.8; }
+
+#registerPackage {
+	margin: 0 auto 0 0;
+	max-width: 30em;
+}
+
+#registerPackage h1 p {
+	font-size: 0.5em;
+}
+
+#registerPackage select,
+#registerPackage input {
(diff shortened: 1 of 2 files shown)


wilzbach:
Btw @ohdatboi if you prefer to use a CSS preprocessor (or a CSS library), don't be afraid to propose it here.
The CSS of the dub-registry just used to be quite small and thus manageable by hand.
See also: https://github.com/dlang/dub-registry/issues/210

ohdatboi:
@wilzbach I would like to use CSS preprocessor. There are a lot of obsolete rules wich forces me to overuse `!important` and duplicate functionality that already exists, but I change CSS very frequently and use of preprocessor will make everything even more complex.
So, CSS should be rewrited in future, but it's better not to mess with it right now.

s-ludwig:
Just one request: The page content should generally left align with the header (Ideally with the "D" of DUB, but that seems to have been broken some time ago).

ohdatboi:
> The page content should generally left align

Done.

> Ideally with the "D" of DUB, but that seems to have been broken some time ago

It's now aligned to the left of the `#content`, so it will be fixed as soon as somebody fixes `#content`'s margins.

s-ludwig:
Okay, thanks!

PR #216, opened 2017-06-08
Fix wrong link to subpackage

Kripth:
The link in the else scope may also be wrong but I couldn't find a package that uses it.


ohdatboi:
Thank you!
Yeah, directory based sub packages should not have links. Could you please fix it?

s-ludwig:
The else branch could in theory just be removed, since path based sub packages are resolved by the registry since a while and appear as normal sub packages in the JSON.

PR #217, opened 2017-06-11
[WIP] Add Gogs support

s-ludwig:
[Gogs](https://gogs.io/) is currently missing API support for querying tags and commit details, so this can't work yet.


s-ludwig:
Note: depends on https://github.com/gogits/gogs/issues/4556

s-ludwig:
Needs to be looked at again if there is interest. We are currently using Gitea after switching to GitLab for a while, so I have no idea whether the Gogs API has the required functionality these days.

PR #218, opened 2017-06-11
[WIP] Add resource builder

ohdatboi:
First step of #210.
Resources are moved into `assets` directory. `resbuild.d` just copies files into `public` dir.


wilzbach:
Hmm while this looks nice in general, is it really worthwhile to start from scratch and not use one of the zillions of existing build tools?

ohdatboi:
I have some experience with Gulp and Rake. They are great tools and I would like to use one of them, but it will require NodeJS/Ruby installed.
Make is everywhere, but it comes with a lot of pain and I don't want to deal with it.

Actually, I just changed my mind. I will use NodeJS/‎NPM/Gulp, because writing own tool is not an answer.
Yeah, it requiers NodeJS installed, but NPM can help with installation of JS/CSS minifiers and other tools. Also, NodeJS is some kind of standard of frontend development.
‎

wilzbach:
> I will use NodeJS/‎NPM/Gulp, because writing own tool is not an answer.

I would recommend having a look at webpack - it's one of the state of the art tool in the fronted Dev world and very convenient to use.

wilzbach:
> From my point of view - main idea of webpack is to manage dependencies required by JS code (like styles, images, etc), which is not what we need in dub-registry.

Yes it is very easy to add any kind of resource (from SASS over to icons) and even generate proper source maps.

> We need a simple builder for styles and scripts.

I thought that's what you were looking for. Generate one blob that contains all the JavaScript and compiled styles? No?

ohdatboi:
@wilzbach I have been confused because of [this](https://webpack.js.org/guides/output-management/), but now I get it. I have to write a plugin which will output simple diet template with all styles/scripts listed. This template will be used in `layout.dt`.

wilzbach:
> Make is everywhere, but it comes with a lot of pain and I don't want to deal with it.

Btw I wouldn't have been a pain with Make, actually quite easy:

```
ASSETS_DIR=assets

################################################################################
# Prerequisites
################################################################################

CSS_FILES=$(addprefix public/styles/, $(addsuffix .css, common markdown top top_p))
FONT_FILES=$(addprefix public/fonts/, fontello.woff)
JS_FILES=$(addprefix public/scripts/, $(addsuffix .js, clipboard.min home menu))
IMG_FILES=$(addprefix public/, favicon.ico $(addprefix images/, clippy.svg \
		  $(addsuffix .png, dub-header dub-logo logo-small)))

################################################################################
# Target directory
################################################################################

JS_BUNDLE=$(ASSETS_DIR)/bundle.js
CSS_BUNDLE=$(ASSETS_DIR)/bundle.css
ASSETS=$(subst public/, $(ASSETS_DIR)/, $(CSS_BUNDLE) $(JS_BUNDLE) $(FONT_FILES) $(IMG_FILES))

all: $(ASSETS) | $(ASSETS_DIR)

$(ASSETS_DIR):
	mkdir -p $@

$(JS_BUNDLE): $(JS_FILES) | $(ASSETS_DIR)
	cat $^ > $@

$(CSS_BUNDLE): $(CSS_FILES) | $(ASSETS_DIR)
	cat $^ > $@

$(ASSETS_DIR)/%: public/% $(ASSETS_DIR)
	mkdir -p $(dir $@)
	cp $< $@
```

ohdatboi:
So, it works! Webpack generates files with unique hashes and `manifest.json`. There is new function `resource` in `layout.dt` which takes resource name, reads `manifest.json` and returns hash.
It re-reads `manifest.json` on every request, but I'll fix it later. I think that it can be readed once at compile time for the `release` build, but at run time for the `debug` build, so I don't have to recompile everything everytime I edit resources.

views/layout.dt
```diff
@@ -1,12 +1,16 @@
 ﻿!!! 5
 html
+	- string resource(string name)
+		- import vibe.data.json;
+		- import std.file : readText;
+		- auto j = "public/manifest.json".readText.parseJsonString;
+		- return j[name].type == Json.Type.undefined ? "" : j[name].get!string;
```
wilzbach: Hmm not so sure whether you want to read the JSON file on every request. A common strategy is to put the build in a unique directory and expose that one as environment variable. Of course, if the variable isn't defined, you could parse the manifest.
I know that with this approach it's harder to switch to new js without restarting the server, but afaict they are currently deployed together anyways​..
Alternatively if that's easier, just read the JSON once on startup.

wilzbach:
> It re-reads manifest.json on every request, but I'll fix it later. I think that it can be readed once at compile time for the release build, but at run time for the debug build, so I don't have to recompile everything everytime I edit resources

I am sorry - I didn't see this. Yes just use `version` and in the release block
use something like `initOnce`

https://dlang.org/library/std/concurrency/init_once.html

.gitignore
```diff
@@ -4,4 +4,5 @@
 /dub-registry
 *.exe
 log.txt
+public/
```
ohdatboi: Already in gitignore

ohdatboi:
@wilzbach Your idea about reading JSON on startup is much better. It solves problem about recompiling program everytime resources are changed and speedups `resource()` function.

I'll fix conflicts. No problem at all.

wilzbach:
> It solves problem about recompiling program everytime resources are changed and speedups resource() function.

If I am not mistaken if you run `webpack -w` (watch mode), the chunk file hash won't change on incremental recompiles. Otherwise I liked the idea of re-reading the manifest in debug mode.

.gitignore
wilzbach: Sorry - reviewed from my phone.

ohdatboi:
> Otherwise I liked the idea of re-reading the manifest in debug mode.

I tried to add global associative array, which will contain pairs of `HASH: string`, but I got errors on userman compilation. `layout.dt` cannot import my module `dubregistry.resource`.

MartinNowak:
What about running the asset compilation as preBuildCommand and then compiling in the digest mapping (e.g. via `parseJSON(import("manifest.json"));` and mixin or by running a small script that generates a D function)?

MartinNowak:
> Looks like userman uses registry's layout.dt, but it doesn't have dubregistry directory in import paths. Well, that sucks.

Yes, that's done so that dependent packages can customize the views, so indeed layout.dt gets compiled into userman as well.

MartinNowak:
Thanks, how about simply using yuglify to generate a single css and js file?
There aren't that many assets and they're changing often either, so there isn't much to save and using webpack looks a bit overblown.
Even for yuglify I'd prefer it to be optional (e.g. let the CI enforce that files are in sync or so) or just use the plain assets in debug mode.
Requiring people to install a JS build tool to use (and develop) dub-registry would be quite a price to pay.

ohdatboi:
> Thanks, how about simply using yuglify to generate a single css and js file?

This [yuglify](https://github.com/yui/yuglify)? It looks dead.
Webpack is a great tool, but I have to find out how to use it in dub-registry. For now I have no idea, so I am thinking about moving to something else. 

> There aren't that many assets and they're changing often either, so there isn't much to save and using webpack looks a bit overblown.

Why webpack? Well, it's an easy question - it is popular. There is no reason to use dead or dying tool. [This](http://stateofjs.com/2016/buildtools/) article tells me that gulp is fine. Not super-popular, but still young and shiny.
So, I think about moving to gulp. I already used it, it has a lot of nice plugins and it is more like general build tools (Make, Rake, etc), than webpack, which is a bundler for JS.

> Requiring people to install a JS build tool to use (and develop) dub-registry would be quite a price to pay.

All you need to use it is npm and NodeJS. npm will install all required dependencies (JS/CSS minimizers, webpack or other tools).

ohdatboi:
So, I used gulp instead of webpack. It can uglify JS and minimize CSS.
`cd assets && npm up` to download all dependencies
`npm run build` - uglifyes JS and minimizes CSS
`npm run watch` - watches for changes, just copies files without uglifying/minimizing them.

PR #223, opened 2017-06-16
support RedHat CA Bundle

MartinNowak:


source/app.d
```diff
@@ -55,8 +57,12 @@ shared static this()
 
 	version (linux) {
 		logInfo("Enforcing certificate trust.");
+		enum debianCA = "/etc/ssl/certs/ca-certificates.crt";
+		enum redhatCA = "/etc/pki/tls/certs/ca-bundle.crt";
+		certPath = redhatCA.exists ? redhatCA : debianCA;
```
MartinNowak: Left the debian CA as default, though we might check existance of both and throw an error if backwards compat is not a concern.

PR #224, opened 2017-06-19
package rank story

MartinNowak:
- compute ranking between 0 and 5
- show ranking only with `?showranking` query parameter for now
- more stat details visible on hover (better visualization/explanation pending)
- sort search results by relevance (rounded ranking and [FTS score](https://docs.mongodb.com/manual/reference/operator/query/text/#text-operator-text-score))

search_ranking (mouse cursor not visible)

Commits are properly structured and should be reviewed one-by-one.

**Note for deployment:** it will take a full update cycle to populate the package stats with real data. Only on the 2nd update round (after ~30min.) we can compute the stat distributions and ratings. So initially all packages will show as equally zero rated which behaves like the status quo ;).

### DIFF
--- a/source/dubregistry/registry.d
+++ b/source/dubregistry/registry.d
@@ -160,7 +160,11 @@ class DubRegistry {
 
 		try {
 			stats.repo = getRepositoryInfo(pack.repository).stats;
-		} catch (Exception e){
+		} catch (FileNotFoundException e) {
+			// repo no longer exists, rate it down to zero (#221)
+			logInfo("Zero rating %s because the repo no longer exists.", packname);
+			stats.rating = 0;
+		} catch (Exception e) {
 			logWarn("Failed to get repository info for %s: %s", packname, e.msg);
 			return typeof(return).init;
 		}
--- a/source/dubregistry/registry.d
+++ b/source/dubregistry/registry.d
@@ -42,12 +42,18 @@ class DubRegistry {
 		PackageWorkQueue m_updateQueue;
 		// list of packages whose statistics need to be updated
 		PackageWorkQueue m_updateStatsQueue;
+		DbStatDistributions m_statDistributions;
 	}
 
 	this(DubRegistrySettings settings)
 	{
 		m_settings = settings;
 		m_db = new DbController(settings.databaseName);
+
+		// recompute ratings on startup to pick up any algorithm changes
+		m_statDistributions = m_db.getStatDistributions();
+		recomputeRatings(m_statDistributions);
+
 		m_updateQueue = new PackageWorkQueue(&updatePackage);
 		m_updateStatsQueue = new PackageWorkQueue((p) { updatePackageStats(p); });
 	}
@@ -85,9 +91,9 @@ class DubRegistry {
 
 	auto searchPackages(string query)
 	{
-		static struct Info { string name; DbPackageVersion _base; alias _base this; }
+		static struct Info { string name; DbPackageStats stats; DbPackageVersion _base; alias _base this; }
 		return m_db.searchPackages(query).filter!(p => p.versions.length > 0).map!(p =>
-			Info(p.name, m_db.getVersionInfo(p.name, p.versions[$ - 1].version_)));
+			Info(p.name, p.stats, m_db.getVersionInfo(p.name, p.versions[$ - 1].version_)));
 	}
 
 	RepositoryInfo getRepositoryInfo(DbRepository repository)
@@ -158,6 +164,11 @@ class DubRegistry {
 			logWarn("Failed to get repository info for %s: %s", packname, e.msg);
 			return typeof(return).init;
 		}
+
+		if (auto pStatDist = pack.repository.kind in m_statDistributions.repos)
+			stats.rating = computeRating(stats, m_statDistributions.downloads, *pStatDist);
+		else
+			logError("Missing stat distribution for %s repositories.", pack.repository.kind);
 
 		m_db.updatePackageStats(pack._id, stats);
 		return stats;
@@ -275,6 +286,8 @@ class DubRegistry {
 	void updatePackages()
 	{
 		logDiagnostic("Triggering package update...");
+		// update stat distributions before rating packages
+		m_statDistributions = m_db.getStatDistributions();
 		foreach (packname; this.availablePackages)
 			triggerPackageUpdate(packname);
 	}
@@ -472,6 +485,18 @@ class DubRegistry {
 
 		m_updateStatsQueue.put(packname);
 	}
+
+	/// recompute all ratings based on cached stats, e.g. after updating algorithm
+	private void recomputeRatings(DbStatDistributions dists)
+	{
+		foreach (packname; this.availablePackages)
+		{
+			const pack = m_db.getPackage(packname);
+			auto stats = m_db.getPackageStats(packname);
+			stats.rating = computeRating(stats, dists.downloads, dists.repos[pack.repository.kind]);
+			m_db.updatePackageStats(pack._id, stats);
+		}
+	}
 }
 
 private PackageVersionInfo getVersionInfo(Repository rep, RefInfo commit, string first_filename_try, Path sub_path = Path("/"))
@@ -541,4 +566,48 @@ struct PackageVersionInfo {
 struct PackageInfo {
 	PackageVersionInfo[] versions;
 	Json info; /// JSON package information, as reported to the client
+}
+
+/// Computes a package rating from given package stats and global distributions of those stats.
+private float computeRating(DownDist, RepoDist)(in ref DbPackageStats stats, DownDist downDist, RepoDist repoDist)
+{
+	import std.math : log1p, round, tanh;
+
+    if (!downDist.total.sum) // no stat distribution yet
+        return 0;
+
+	/// Using monthly downloads to penalize stale packages, logarithm to
+	/// offset exponential distribution, and tanh as smooth limiter to [0..1].
+	immutable downloadRating = tanh(log1p(stats.downloads.monthly / downDist.monthly.mean));
+	logDebug("downloadRating %s %s %s", downloadRating, stats.downloads.monthly, downDist.monthly.mean);
+
+	// Compute rating for repo
+	float sum=0, wsum=0;
+	void add(T)(float weight, float value, T dist)
+	{
+		if (dist.sum == 0)
+			return; // ignore metrics missing for that repository kind
+		sum += weight * log1p(value / dist.mean);
+		wsum += weight;
+	}
+	with (stats.repo)
+	{
+		alias d = repoDist;
+		// all of those values are highly correlated
+		add(1.0f, stars, d.stars);
+		add(1.0f, watchers, d.watchers);
+		add(1.0f, forks, d.forks);
+		add(-1.0f, issues, d.issues); // penalize many open issues/PRs
+	}
+
+	immutable repoRating = max(0.0, tanh(sum / wsum));
+	logDebug("repoRating: %s %s %s", repoRating, sum, wsum);
+
+	// average ratings
+	immutable avgRating = (repoRating + downloadRating) / 2;
+	assert(0 <= avgRating && avgRating <= 1.0, "%s %s".format(repoRating, downloadRating));
+	immutable scaled = stats.minRating + avgRating * (stats.maxRating - stats.minRating);
+	logDebug("rating: %s %s %s %s %s %s", stats.downloads.monthly, downDist.monthly.mean, downloadRating, repoRating, avgRating, scaled);
+
+	return scaled;
 }
--- a/views/view_package.dt
+++ b/views/view_package.dt
@@ -152,26 +152,31 @@ block body
 					a.blind(href="#{req.rootDir}packages/#{urlEncode(packageName)}/versions")
 						| Show all #{packageInfo["versions"].length} versions
 
-			dt Stats:
+			- auto stats = registry.getPackageStats(packageName);
+
+			dt Download Stats:
 			dd#stats
-				- auto stats = registry.getPackageStats(packageName).downloads;
 				ul.unmarkedList
 					li
 						p
-							strong= stats.daily.to!string
+							strong= stats.downloads.daily
 							|  downloads today
 					li
 						p
-							strong= stats.weekly.to!string
+							strong= stats.downloads.weekly
 							|  downloads this week
 					li
 						p
-							strong= stats.monthly.to!string
+							strong= stats.downloads.monthly
 							|  downloads this month
 					li
 						p
-							strong= stats.total.to!string
+							strong= stats.downloads.total
 							|  downloads total
+
+			- if ("showrating" in req.query)
+				dt Rating:
+				dd#rating(title="#{formatPackageStats(stats)}")= formatRating(stats.rating)
 
 	script(type="application/javascript", src="/scripts/clipboard.min.js")
 	:javascript
--- a/source/dubregistry/dbcontroller.d
+++ b/source/dubregistry/dbcontroller.d
@@ -63,6 +63,7 @@ class DbController {
 
 		// create indices
 		m_packages.ensureIndex([tuple("name", 1)], IndexFlags.Unique);
+		m_packages.ensureIndex([tuple("stats.rating", 1)]);
 		m_downloads.ensureIndex([tuple("package", 1), tuple("version", 1)]);
 
 		Bson[string] doc;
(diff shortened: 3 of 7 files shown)


views/view_package.dt
```diff
@@ -98,12 +98,12 @@ block body
 
 		- if( auto pa = "authors" in versionInfo )
 			p#authors.wrap
-				strong Authors:
```
ohdatboi: I know, trailing whitespaces is a bad thing, but this one is necessary. It adds spacing after `Authors:`.
Also, space after `Sub packages:`, `Dependencies:` and `System dependencies:` is also required.

That's why it is better to keep coding style fixes and new features in separate PRs.

ohdatboi:
I like it! But I think that repository stats should be displayed on the package page. Tooltip is very big and not mobile-friendly. 
It can be displayed in `Stats` section, after (or before) downloads.

MartinNowak:
> It can be displayed in Stats section, after (or before) downloads.

Yes, should be there as well.

views/view_package.dt
MartinNowak: That was my editor, didn't notice, thanks.

s-ludwig:
I think we should give this a unique name (similar to "Google page rank"), since it is quite an abstract measure and the current impression is more of a user rating, which gives lower scored projects an unfair disadvantage. For the latter reason it seems to me that we should either not display the actual value at all (just use it for sorting), or at least in a way that doesn't resemble a user rating. Libraries.io has a similar concept ("SourceRank") and simply displays it as a numeric value (e.g. https://libraries.io/dub/vibe-d)

MartinNowak:
OK, will experiment with an unbounded numeric value instead.

source/dubregistry/registry.d
```diff
@@ -575,4 +628,48 @@ struct PackageVersionInfo {
...
+	immutable repoRating = max(0.0, tanh(sum / wsum));
+	logDebug("repoRating: %s %s %s", repoRating, sum, wsum);
+
+	// average ratings
+	immutable avgRating = (repoRating + downloadRating) / 2;
+	assert(0 <= avgRating && avgRating <= 1.0, "%s %s".format(repoRating, downloadRating));
+	immutable scaled = stats.minRating + avgRating * (stats.maxRating - stats.minRating);
+	logDebug("rating: %s %s %s %s %s %s", stats.downloads.monthly, downDist.monthly.mean, downloadRating, repoRating, avgRating, scaled);
+
+	return scaled;
```
andralex: Cool, would be great to add a comment that summarizes the computation in somewhat math/natural language notation, e.g. "weighted sum of stars, watches, forks with all weights 1.0, minus penalty for issues weighted by 1.0 each" etc.

I assume there will be some tweaking over time.

source/dubregistry/repositories/gitlab.d
```diff
@@ -1,120 +1,121 @@
...
+class GitLabRepository : Repository {
+	private {
+		string m_owner;
+		string m_project;
+		URL m_baseURL;
+		string m_authToken;
+	}
+
+	static void register(string auth_token, string url)
+	{
```
andralex: Some code here has Egyptian braces, some other has Allman braces.

andralex: Lots of tabs too :)

s-ludwig:
@MartinNowak BTW, maybe we could also just add a hidden parameter that enables the star rating, so that we can evaluate the rating system on the live system, without normal users seeing this yet.

source/dubregistry/repositories/gitlab.d
ZombineDev: Dub and vibe.d follow a different style than Phobos, see: https://vibed.org/style-guide.

s-ludwig: Basic rationale: The idea is to optimize for readability. Usually avoiding to waste vertical space is important to improve the visual structure and the amount of code visible at a time. This is where Egyptian braces are good. For function signatures the situation is a bit different, because it's important that they stand out while skipping through a document. Having them separated by an almost blank line can help a lot there. It also works better with multi-line declarations.

andralex: By that argument struct and class definitions are also important... anyhow the larger point is "convention" and "conventional" have the same root, which makes an unconventional convention (as this seems to be) stand out oddly. Why not go with something that is so established it even has a name?

s-ludwig: > By that argument struct and class definitions are also important

This is true to some extent, and I'd probably do it differently there today. The reason it is that way is that the style has been adjusted a few times in the past (mostly to follow language and Phobos style changes).

> anyhow the larger point is "convention" and "conventional" have the same root, which makes an unconventional convention (as this seems to be) stand out oddly. Why not go with something that is so established it even has a name?

This originates from our internal code base. Whatever the style used there is *is the* convention, there is very little reason to care what is used in other groups. And of course I'm using the style that I'm comfortable with for my open source projects, too, especially since they are usually born out of an internal project.

Regarding named "standards", I don't think it makes a big difference in practice if a convention has a name or not. Most people will have to learn the details no matter what, unless everything happens within the same organization/community. I'm all for adapting existing conventions when they are good, but I hate when it's applied dogmatically and gets in the way of better solutions (speaking generally).

Anyway, if you want to change everything DUB to be in Phobos style, I won't object., even if I highly dislike the brace style within function bodies (and I won't start with tabs vs. spaces :wink:).

andralex: >Regarding named "standards", I don't think it makes a big difference in practice if a convention has a name or not.

I agree it doesn't make a big difference. It makes a HUGE difference. "The moment you need to explain a convention, you've lost."

ohdatboi: @andralex There are not a lot of contributors, and forcing them to follow some kind of convention is just a waste of their time. Also, dub-registry is full of Diet templates, CSS styles and JS scripts, so convention should cover them too, which requires a lot of work.

s-ludwig: >> Regarding named "standards", I don't think it makes a big difference in practice if a convention has a name or not.
>
> I agree it doesn't make a big difference. It makes a HUGE difference. "The moment you need to explain a convention, you've lost."

Erm, except that you have to explain the named conventions as well.

andralex:
@ohdatboy: but a convention does exist, it's just not conventional and that
makes it wrong.

@s-ludwig: next time we meet let's discuss visitor, except every time I
ought to mention it I'll insert its description.

I'm shocked that this conversation exists.
-- 
Please excuse terseness and typos

s-ludwig:
> next time we meet let's discuss visitor, except every time I
> ought to mention it I'll insert its description.
>
> I'm shocked that this conversation exists.

We must be missing some point of each other. Let me put it this way, if I don't know what a visitor pattern is, the name won't help me. There are countless (even popular) code style conventions out there, so that using one of them probably won't help most people. Also, w.r.t. to the "vibe.d code style", please compare against the most well known style that I can think of: https://en.wikipedia.org/wiki/Indent_style#K.26R

CyberShadow:
FWIW, one (very small) reason I haven't contributed much to Vibe and Dub is the code style. I write all my personal code in Phobos style (except I use tabs, but with a well configured editor there is no difference), and a lot of code in the D ecosystem (at least with which I had to work with) uses Phobos style as well. Having to read and write code that uses an unusual style creates a little bit of friction.

MartinNowak:
> @MartinNowak BTW, maybe we could also just add a hidden parameter that enables the star rating, so that we can evaluate the rating system on the live system, without normal users seeing this yet.

I think you were right with the "it looks too much like a user rating" point, because of the x of 5.0, let me try an unlimited (no tanh) ranking.

I'd prefer if noone hijacked PRs for style discussion. Adopt and maintain [dfmt](https://github.com/dlang-community/dfmt) and stop wasting our time.

source/dubregistry/registry.d
```diff
@@ -575,4 +628,48 @@ struct PackageVersionInfo {
...
+/// Computes a package rating from given package stats and global distributions of those stats.
+private float computeRating(DownDist, RepoDist)(in ref DbPackageStats stats, DownDist downDist, RepoDist repoDist)
+{
+	import std.math : log1p, round, tanh;
+
+    if (!downDist.total.sum) // no stat distribution yet
+        return 0;
+
+	/// Using monthly downloads to penalize stale packages, logarithm to
+	/// offset exponential distribution, and tanh as smooth limiter to [0..1].
```
MartinNowak: not a noun
https://en.wikipedia.org/wiki/Limiter

CyberShadow:
> I'd prefer if noone hijacked PRs for style discussion.

Sorry. I'll delete my comment.

wilzbach:
Alrighty - I finally got around deploying this to http://alpha.dub.pm/

@MartinNowak @s-ludwig @CyberShadow  I have added your keys to the `root` user, s.t. you can do use this machine as well if you want to. It's is a very cheap VPS (15$/year) and it will be terminated at the end of this year. I have simply used `screen` to start this test process.

s-ludwig:
I'm seeing this, looks like it still crashes during the score computation: 
image

Is there any stack trace in the log? (actually tried to SSH into the machine myself, but the default port appears closed)?

wilzbach:
> Is there any stack trace in the log?

That's really weird. It ran fine when I uploaded it. I built it with `ldc` 1.3.0 in release mode (on my machine as the VPS isn't powerful enough to build the dub-registry and I didn't want to wait for the separate build).

I just checked - it didn't crash and it also updates the stats for packages:

```
Updating stats for pegged
Updating tags for sdlang-d: ["v0.8.1", "v0.8.2", "v0.8.3", "v0.8.4", "v0.9.0", "v0.9.1", "v0.9.2", "v0.9.3", "v0.9.4", "v0.9.5", "v0.9.6", "v0.10.0", "v0.10.1"]
Error for version 0.8.1 of sdlang-d: Found no package description file in the repository.
Error for version 0.8.2 of sdlang-d: Configuration is missing a name.
Error for version 0.8.3 of sdlang-d: Configuration is missing a name.
Updating branches for sdlang-d: ["schema", "master"]
Updating stats for sdlang-d
Updating tags for llvm-d: ["v1.0.0", "v1.1.0", "v1.1.1", "v2.0.0", "v2.0.1", "v2.0.2", "v2.0.3", "v2.1.0", "v2.1.1", "v2.2.0"]
Updating branches for llvm-d: ["master"]
Updating stats for llvm-d
Updating tags for gl3n: ["v1.0.0", "v1.0.1", "v1.1.0", "v1.2.0", "v1.3.0", "v1.3.1"]
Updating branches for gl3n: ["master", "gh-pages"]
Error for branch ~gh-pages of gl3n: Found no package description file in the repository.
Updating stats for gl3n
```

However now only one package is shown:

image

>  (actually tried to SSH into the machine myself, but the default port appears closed)?

I completely forgot (I use shortcuts, s.t. I don't have to enter this) - the port is `13932`

s-ludwig:
I'm going to remove the package statistics display on master until this is ready, because it really makes the registry almost unusable.

andralex:
Folks, what is the status of this? Don't forget... important and urgent. Progress in the matter would be fantastic. Thanks!

MartinNowak:
I've restructured the PR to separate out the stats gathering and caching, see #240.

andralex:
Can't wait to see this in! @s-ludwig are you the sole reviewer?

s-ludwig:
Merged #240.

@andralex: @wilzbach and @CyberShadow were also involved in our initial package scoring kick-off at DConf, so I'd expect them to chime in for the high-level issues, too, if they feel like they need to.

CyberShadow:
My contribution was solely limited to some brain storming... considering I'm not very familiar or involved with Dub and its architecture, I don't see how I can meaningfully contribute here.

s-ludwig:
@andralex What exactly do you expect in terms of review right now? I expressed my concerns with the general rating scheme, or rather its current visual representation. The rest of the changes has been merged as #240. What we need now is to find a good representation that doesn't stimulate the wrong associations (e.g. quality vs popularity).

ohdatboi:
> What we need now is to find a good representation that doesn't stimulate the wrong associations (e.g. quality vs popularity).

Are we talking about rank system, which is used to build lists of best packages and sort search results, or about information showed to user on package page? In second case we could just display:
- **NEW** badge on new packages (Created less than a week ago)
- **No. 42 Most Downloaded** - package position on the list of the most downloaded packages.
- **No. 13 Hot** :fire: - package position on the list of hot (trending) packages.
- **No. 33 Best** - package position on the list of the best packages.

All of those should be shown only for packages in top 50 of their respective lists, so **No. 999 Best** would never happen in registry of 1000 packages.

It could be added right after package version here:
badges
I know, colours are terrible, but styles for creating such badges are already here, so they could be added in one short commit (Talking about front-end).
They could be added to the side bar as well.

andralex:
@s-ludwig no expectations wrt review, just want to see the feature in as soon as possible. Something imperfect that can be learned from and tweaked later is infinitely better than the current situation. Thanks!

s-ludwig:
I don't think we are at the point yet, where we profit from public experiences, we should work out the most obvious issues first. But one thing we could already extract from the change set would be using the current score formula for sorting the package list by popularity. My main problem is just with the current "star rating" appearance.

@ohdatboi I'd use the term "popularity" for this particular score. It would definitely fit one of the possible top 10 lists, as well as being used to sort search results (combined with the current text match score). I'm not sure yet if it makes sense to actually display it directly to the user as some abstract metric, or if we should just show the underlying numbers on the package page (which we should do anyway).

The badge idea sounds good. I would try to keep the number of badges down, maybe "new", "nth most popular" and a "high quality" badge. The quality badge would be added if some preconditions are true, such as has README, has CI and maybe coverage, last version is newer than half a year and compiles on the latest DMD. There could also be multiple quality levels

ohdatboi:
> I'd use the term "popularity" for this particular score. It would definitely fit one of the possible top 10 lists, as well as being used to sort search results (combined with the current text match score). I'm not sure yet if it makes sense to actually display it directly to the user as some abstract metric, or if we should just show the underlying numbers on the package page (which we should do anyway).

I think it shouldn't be displayed, because it is an abstract number and people will get it wrong. If popularity/score/rank is in range 0-10, people will get it as basic "something out of 10" thing,  if it is in range 0-100, they will think about percents, and so on. Using bounded value is bad, but using an abstract number is even worse. So, not showing it at all is the best chance.

> I would try to keep the number of badges down, maybe "new", "nth most popular" and a "high quality" badge.

Picture is just an example, since package cannot have "new" and "most downloaded" at the same time or "new" and "best", so vibe-d, for example, will have only 3 badges (best, most downloaded, hot). Also, carefully chosen colours would make it fit current style.

Now I am thinking that "hot" is not a good badge, since it will be pretty much same as the "best", so it can be removed. And "best" also makes no sense, since it is pretty much what "most downloaded" means.

> The quality badge would be added if some preconditions are true, such as has README, has CI and maybe coverage, last version is newer than half a year and compiles on the latest DMD. There could also be multiple quality levels

Nice idea! I don't know if it is hard to implement such a thing, but I can draw a badge.

MartinNowak:
Yes, let's stick with the terminology "ranking", instead of "rating", and similarly use numbers instead of stars. FWIW my implemenation works a lot like https://libraries.io/dub/derelict-gl3/sourcerank which looks reasonable.

Sorting packages without displaying why sth. is sorted higher sounds equally flawed. I don't think our audience has much problems with numbers.
I don't like the badges idea too much, as it adds several new but similar dimensions. Most of those metrics should become sortable in search results anyhow, so badges don't even add much functionality.

We've discussed a `top10, trending10, newest10/random10` triptych for the landing page though.
https://dlang.slack.com/archives/C3P93MVU6/p1488125367199441

@s-ludwig will you deploy #240 soon, it's kind of required to continue working on this as the local download stats are already outdated again. Having the stats be mirrored would help a lot.

ohdatboi:
> Sorting packages without displaying why sth. is sorted higher sounds equally flawed.

That's how search on various sites works today. Google search, search on NPM or any other site sorts results without showing any kind of rank/rating.

>  I don't think our audience has much problems with numbers.

Yep, not a big problem for people with mathematics degree, but it'd confuse others. What is this number? Why is it here? Is X enough? How do I increase it? Should I increase it? Is it important?

That's why shiny, colourful badges are better than a dull number.

> I don't like the badges idea too much, as it adds several new but similar dimensions.

They're just a small front-end enhancement, super-easy to implement (If related lists are implemented), but useful.

> Most of those metrics should become sortable in search results anyhow, so badges don't even add much functionality.

Badges display position of the package on the global lists. Those lists should be implemented later, and they have nothing to do with ranking system. Looks like this discussion should move somewhere else (slack channel, for example)

MartinNowak:
Also without visible numbers help to get feedback for/debug the algorithm.

> What is this number? Why is it here? Is X enough? How do I increase it? Should I increase it? Is it important?

It's called ranking, should be fairly obvious. Also there is/will be a breakdown of how this was computed. It seems that the badge discussion is OT when it's not intended to be used for sorting.

s-ludwig:
> Also without visible numbers help to get feedback for/debug the algorithm.

This is a good point, but we can hide that behind an explicit step to enable debug information, maybe a small link at the bottom that appends "&scoring_debug=1" or similar.

MartinNowak:
Updated now with a hidden-by-default implementation, this should allow us to move on with this and improve the details step by step.

PR #228, opened 2017-06-29
New layout of login page

ohdatboi:
Same as #215 but for login page.
login

### DIFF
--- a/public/styles/common.css
+++ b/public/styles/common.css
@@ -351,17 +351,17 @@ ul.unmarkedList > li:hover,
 
 .light { opacity: 0.8; }
 
-#registerPackage {
+#inputForm {
 	margin: 0 auto 0 0;
 	max-width: 30em;
 }
 
-#registerPackage h1 p {
+#inputForm h1 p {
 	font-size: 0.5em;
 }
 
-#registerPackage select,
-#registerPackage input {
+#inputForm select,
+#inputForm input {
 	box-sizing: border-box;
 	width: 100%;
 	padding: 1em;
@@ -370,15 +370,15 @@ ul.unmarkedList > li:hover,
 	border-radius: 2px;
 }
 
-#registerPackage label {
+#inputForm label {
 	font-weight: bold;
 }
 
-#registerPackage form p {
+#inputForm form p {
 	margin-bottom: 1em;
 }
 
-#registerPackage button {
+#inputForm button {
 	padding: 1em;
 	margin-top: 1em;
 	color: #fff;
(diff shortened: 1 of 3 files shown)


ohdatboi:
Fixed typos.
Screenshot:
loginfixed

public/styles/common.css
```diff
@@ -351,17 +351,17 @@ ul.unmarkedList > li:hover,
 
 .light { opacity: 0.8; }
 
-#registerPackage {
+#inputForm {
```
MartinNowak: Make it `.inputForm` then, no?

MartinNowak:
We might want to evaluate using some simple CSS framework, such as http://getskeleton.com/, instead of writing everything from the ground up. No opinion on how important look-alike to dlang.org is though.

public/styles/common.css
ohdatboi: There are no cases with more than 1 input form on the page, but yes, making it `.inputForm` looks better. I'll fix it.

ohdatboi:
@MartinNowak I was thinking about using CSS preprocessor (like SASS). This requires complete rewriting and rethinking styles, cleaning dead/unused rules, etc.
I would like to use some kind of framework, but framework *and* preprocessor are much better (avoiding copy/paste, adding some useful variables and more).

Anyway, there are not a lot of responses in #218, which adds resource builder. It should have more opinions, so bad/useless stuff won't get merged into master.

wilzbach:
Currently:

image

New:

image

wilzbach:
@s-ludwig 
1) Can we merge such trivial stuff without your approval?
2) Is there a trivial way to setup auto-deploy for the DUB registry?

s-ludwig:
> Can we merge (or squash merge in this case) such trivial stuff without your approval?

No problem. I actually had this stored in my mind as merged because Martin had already approved it.

> Is there a trivial way to setup auto-deploy for the DUB registry?

We'd at least need to have a quick way to roll back changes in case of any regressions, which probably wouldn't qualify as trivial anymore. But I realized that I really need to invest some time into a better deployment and service management system anyway (better crash reports and redundancy to hide the effects of intermittent failures would also help a lot). I'll try to reserve some time for this next week.

wilzbach:
> We'd at least need to have a quick way to roll back changes in case of any regressions, which probably wouldn't qualify as trivial anymore. 

Depending on the deployment time, it might be feasible to simply revert the PR and send it to the deployment queue.

> But I realized that I really need to invest some time into a better deployment and service management system anyway (better crash reports and redundancy to hide the effects of intermittent failures would also help a lot). I'll try to reserve some time for this next week.

If you use sth. like `ansible` and put your scripts here, I could probably hook the VPS, s.t. it's `staging.dub.pm`, i.e. we directly deploy all commits to master on it, but we only do releases whenever someone has time to look out in case of failures?

PR #232, opened 2017-07-20
New layout of the register page

ohdatboi:
Changes layout of the register page.

Screenshots:
register
register_error

### DIFF
--- a/views/userman.register.dt
+++ b/views/userman.register.dt
@@ -1,0 +1,35 @@
+extends layout
+
+block title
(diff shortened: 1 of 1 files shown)


s-ludwig:
At some point we should probably look into unifying this with userman, so that this becomes just a matter of different CSS styling. The current table based layout is not a good default obviously. But independently of that, this is definitely nice progress on the dub-registry side and shouldn't be blocked by it.

PR #238, opened 2017-07-25
New layout of the profile page

ohdatboi:
Changes layout of the profile page.

profile

### DIFF
--- a/views/userman.profile.dt
+++ b/views/userman.profile.dt
@@ -11,7 +11,7 @@ block body
 			- if (useUserNames)
 				p
 					label(for="user_name") User name
-					input(type="text", name="user_name", readOnly, value=user.name)
+					input(type="text", name="name", readOnly, value=user.name)
 			p
 				label(for="full_name") Display name
 				input(type="text", name="full_name", value=req.form["full_name"])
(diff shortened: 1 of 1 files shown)


views/userman.profile.dt
```diff
@@ -0,0 +1,34 @@
...
+
+block body
+
+	.inputForm
+		h1 Personal information
+		form(action="profile", method="POST")
+			- if (useUserNames)
+				p
+					label(for="user_name") User name
+					input(type="text", name="user_name", readOnly, value=user.name)
```
s-ludwig: Not that it matters much due to its read-only state, but this should be just `name="name"` according to the userman implementation (it allows changing the user name in theory, but there is currently a bug in the web code).

s-ludwig:
Actually, I just used the handy online editor ;)

Thanks for doing this! The whole site slowly starts looking like a modern service. I hope we'll have a good rating system soon, so that we can start to design a clean front page and a new search result style.

ohdatboi:
@s-ludwig Yep, all things related to registration and profile management are done. Now I am going to work on `Edit package` page.

PR #241, opened 2017-08-08
[WIP] Refactoring stylesheets

ohdatboi:
I'm trying to get rid of unused and obsolete rules in stylesheets and refactor them. This will improve stylesheet maintainability.

Here is the plan:
- [ ] Get rid of unused/obsolete rules
- [ ] Merge rules defined in multiple files
- [ ] Add useful comments


MartinNowak:
> I'm trying to get rid of unused and obsolete rules in stylesheets and refactor them. This will improve stylesheet maintainability.

Are you aware that this is based on [dlang.org's stylesheet](https://github.com/dlang/dlang.org/blob/ba9cd9e5979cbe9d73ce7e6883cc5a5f5df26d6c/css/style.css)?
I'm not against forking this as we're heading somewhat towards a remake of various elements (e.g. drop most of the menu) on code.dlang.org, and the 2 sites don't necessarily need to look identical.
Still it should be a concious decision to depart with that stylesheet.

You happen to know [uncss](https://github.com/giakki/uncss)? It's quite a useful tool to strip css rules, just pass it all different subpages (and media queries) to collect everything that is used.

ohdatboi:
> Are you aware that this is based on dlang.org's stylesheet?

Yes, I am. It says `Copied __without__ modifications from dlang.org` at the beginning of the file, even though it has been modified.

> I'm not against forking this as we're heading somewhat towards a remake of various elements (e.g. drop most of the menu) on code.dlang.org, and the 2 sites don't necessarily need to look identical.

`top.css` has been created at Mar 17, 2016 and edited a couple of times, which means that we are already dealing with the fork here.

> Still it should be a concious decision to depart with that stylesheet.

Thing is, this stylesheet is a part of dlang.org and it contains not just general styles, like menus or colours, but specific, used only at dlang.org, which means, we can't just copy/paste it and sync with an original one, because it can be changed in the next commit, breaking every site depending on it.
Dlang-tour, for example, uses only needed parts of `top.css` (about 200 LoC). 

I think that `top.css` should be cleaned up and renamed. `dstyle.css` or `dlang.css`, maybe?

> You happen to know uncss? It's quite a useful tool to strip css rules, just pass it all different subpages (and media queries) to collect everything that is used.

Nice! I'll try it.

wilzbach:
> Dlang-tour, for example, uses only needed parts of top.css (about 200 LoC).

Ideally dlang.org should use minimal stylesheets as well. Maybe you have time to do sth like:

https://github.com/dlang/dlang.org/pull/1318

ohdatboi:
> Ideally dlang.org should use minimal stylesheets as well. 

Good job, but it is impossible to split current `style.css`.
For example:
```css
#top > .helper,
body > .container,
body#Home #content #tools > div,
body#Home #content > .intro > div,
body#Home #content > .more
{
    margin: auto;
    max-width: 76em;
    padding: 0 1em;
}
```
Is just broken. It should be rewrited and SASS won't help here.

> Maybe you have time to do sth like

I want to remove all unnecessary stuff from `top.css`, rename and just use it. Some styles could be rewritten, like menu (Is it necessary to have so much wrapper divs?).

MartinNowak:
> Is just broken. It should be rewrited and SASS won't help here.

I'd be hesitant to introduce more tooling (e.g. SASS). While those are standard frontend tools, we have lots of C++-like folks maintaining the website. Any additional hurdle might get in the way. @wilzbach 
If it's just for variables I'd prefer a low-tech `COLOR_RED` + `sed -i 's/COLOR_RED/#FF0000/g/'` over sass.

ohdatboi:
> If it's just for variables I'd prefer a low-tech COLOR_RED + sed -i 's/COLOR_RED/#FF0000/g/' over sass.

Should I even say how bad is this idea? Debugging such code is much harder than using npm to download and build everything.

Thinking "we doesn't need X, it's an another dependency, we could just write some kind of preprocessor" is wrong. They're ok for small projects, but bigger project gets, harder it is to maintain such things.

That's kind of funny: people, who contribute to "Better C++", are against "Better CSS".

> I'd be hesitant to introduce more tooling (e.g. SASS). While those are standard frontend tools, we have lots of C++-like folks maintaining the website. 

I don't think that CSS is a big problem for dlang.org, but usage of DDOC is. Instead of using some HTML template and articles written in Markdown, they're using DDOC. DDOC is a nice tool for adding some marcos to make things easier, but not for building websites.

SASS and other tools in the dub registry should be used, and I had a PR, but realized, that it's impossible to fix things by adding more complexity. So, I decided to strip all unneeded stuff first.
After all unnecessary rules are deleted from `top.css`, it would be about 600 lines, which is much easier to work with.

s-ludwig:
BTW, I'd still like to find the time some day to separate the DUB specific parts of the site into a separate website, so that only the registry part is left here. That would mean that we could probably get rid of the menu altogether and focus on a simplified style. It would make sense to defer any further cleanup efforts to there.

wilzbach:
> I'd be hesitant to introduce more tooling (e.g. SASS). While those are standard frontend tools, we have lots of C++-like folks maintaining the website. 

OT: That's not entirely true, the people maintaining the website would love to use modern, standard tools and get rid of the non-standard highly complicated Ddoc macros, but that's blocked by A&W.
FWIW: Even the dlang.org template was written in SASS, but A&W vetoed that and thus the original creator left. Luckily anonymous converted it to CSS and pushed it through the review process:

https://github.com/dlang/dlang.org/pull/1190
http://forum.dlang.org/post/n6pdes$1upn$1@digitalmars.com

PR #244, opened 2017-08-14
Fix labels issue in Firefox

ohdatboi:
Fixes this:
fixff

Works for all input forms with `.inputForm` class.


MartinNowak:
Turns out that justified text is rather dangerous as (almost) global.

PR #246, opened 2017-08-21
improve styling for package info

MartinNowak:
- use definition list for named sections and add a bit of spacing
- align all entries to 0.5em padding

### old
package_info_old
### new
package_info_new

### DIFF
--- a/public/styles/common.css
+++ b/public/styles/common.css
@@ -285,7 +285,6 @@ div#versions > p {
 	margin-bottom: 0.5em !important;
 }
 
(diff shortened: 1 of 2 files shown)


s-ludwig:
Would be good to increase the top margin of the titles to around 1.5x to get better visual association with the content below. The left margin of the contents could then maybe also removed to keep a clear vertical line, although it doesn't look particularly disrupting in the pictures.

MartinNowak:
Done @s-ludwig, also updated the screenshot above.

PR #250, opened 2017-08-29
New layout of the "My packages" page

ohdatboi:
Replaces table with a list, so it is mobile-friendly and looks better.

Also, this is what search results page can look like, so any feedback is highly appreciated.

## Screenshots
mypacks_empty
mypacks

### DIFF
--- a/public/styles/common.css
+++ b/public/styles/common.css
@@ -416,3 +416,17 @@ div.packageInfo > dl > dd {
 	color: #a94442;
 	border: 1px solid #ebccd1;
 }
+
+.packageList .row {
+	padding: 1em 0;
+	border-bottom: 1px solid rgba(0,0,0,0.1);
+}
+
+.packageList .row a.name {
+	font-size: 1.5em;
+}
+
+.packageList .row a.name .badge {
+	font-size: 60%;
+	vertical-align: middle;
+}
--- a/public/styles/common.css
+++ b/public/styles/common.css
@@ -416,17 +416,3 @@ div.packageInfo > dl > dd {
 	color: #a94442;
 	border: 1px solid #ebccd1;
 }
-
-.packageList .row {
-	padding: 1em 0;
-	border-bottom: 1px solid rgba(0,0,0,0.1);
-}
-
-.packageList .row a.name {
-	font-size: 1.5em;
-}
-
-.packageList .row a.name .badge {
-	font-size: 60%;
-	vertical-align: middle;
-}
--- a/public/styles/common.css
+++ b/public/styles/common.css
@@ -416,3 +416,34 @@ div.packageInfo > dl > dd {
 	color: #a94442;
 	border: 1px solid #ebccd1;
 }
+
+.packageList .error {
+	color: #a94442;
+}
+
+@media (max-width: 992px) {
+	.packageList .row {
+		padding: 1em 0;
+		border-bottom: 1px solid rgba(0,0,0,0.1);
+	}
+
+	.packageList .row .cell:first-child {
+		font-size: 1.3em;
+	}
+}
+
+@media (min-width: 992px) {
+	.packageList { display: table; }
+	.packageList .row {
+		display: table-row;
+	}
+	.packageList .row .cell:first-child {
+		padding-right: 1em;
+		white-space: nowrap;
+	}
+	.packageList .row .cell {
+		display: table-cell;
+		padding: 0.3em 0;
+		border-bottom: 1px solid rgba(0,0,0,0.1);
+	}
+}
--- a/views/my_packages.dt
+++ b/views/my_packages.dt
@@ -20,18 +20,22 @@ block body
 
 	- else
 		h1 #{user.name}'s packages
-		.inputForm
+		.inputForm(style="padding-bottom: 1em")
 			form(method="GET", action="#{req.rootDir}register_package")
 				button(type="submit") Register new package
-		.packageList
-			- foreach (p; sort!((a, b) => a < b)(packages.array))
-				- auto pack = registry.getPackageInfo(p, true).info;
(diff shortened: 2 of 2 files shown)


ohdatboi:
I've added a heading:
mypacks_heading

s-ludwig:
It would be good to have the description in the same block as the package name/version if the screen is sufficiently large. Otherwise, even though this looks a bit nicer, it would mean that there would potentially be a lot of additional scrolling going on (I have over 30 packages registered for example).

ohdatboi:
> It would be good to have the description in the same block as the package name/version if the screen is sufficiently large.

Every list item should have some empty space around it, so it wouldn't look "heavy". It is impossible to make it nice *and* compact at the same time.

That's why tables aren't used anymore - it is all about displaying information nice and easy, not about displaying as much of it as possible.

> Otherwise, even though this looks a bit nicer,  it would mean that there would potentially be a lot of additional scrolling going on

It can be split into pages, just like on the main page.

s-ludwig:
The problem is that this is a usability loss. Visuals IMO always need to be prioritized *after* usability has been optimized. There is no point making this nice and shiny if the result is that using it gets more cumbersome. There is lots of horizontal space on a monitor (which is definitely the primary device to display these pages, mobile should be the exception), so we should make use of it.

s-ludwig:
Just a quick mock:
image

It looks more condensed, but I don't see that as an issue at all, considering that this is an internal page that is only meant to quickly access personal packages.

s-ludwig:
The errors could also be emphasized a little bit, e.g.:
image

ohdatboi:
I don't think that forcing user to scroll through a nice, shiny list a little bit more is a problem. Menu inside a menu inside a menu is cumbersome, but scrolling isn't.

picturejklhjlhkjlh

It is much harder to read, because it isn't aligned properly (red line). Grey-coloured descriptions are just a noise, and it isn't easy to read them.

Error looks bad because of border. IMO even plain red text looks better.

Using emphasis and icon is a good idea, I'd like to use it.

s-ludwig:
> It is much harder to read, because it isn't aligned properly (red line). Grey-coloured descriptions are just a noise, and it isn't easy to read them.

I bet that reading the version with more vertical spacing, which has no alignment either, is even harder/slower to read.

> Error looks bad because of border. IMO even plain red text looks better.

I'd agree there, it was really just a very quick adaption of your version without putting any though into adjusting it for the new layout.

> Menu inside a menu inside a menu is cumbersome, but scrolling isn't.

I disagree here. Right now I almost have all of my packages visible on a single screen. A quick turn on the scrolling wheel gives me access to the rest. With the vertically expanded layout, I'd have to reach for the scrollbar, click on a page number, or scroll lots of times with the mouse wheel. At the same time I'd have to scan a much larger area to find the package that I'm looking for (while scrolling).

This wouldn't be the end of the world, but I don't see the point of making this worse for a page that really only has utility value. It should be consistent with the rest of the site, but it's not a presentation kind of page where each item needs space to come into its own.

s-ludwig:
> Grey-coloured descriptions are just a noise, and it isn't easy to read them.

The exact font can be argued about, but the description itself is really not a primary piece of information here. It can be useful to spot any spelling errors when one is looking at a particular item, but otherwise it could theoretically as well be dropped completely.

s-ludwig:
BTW, I didn't mean to put out those drafts as necessarily concrete implementation suggestions, there may be all kinds of improvements to achieve a similar compactness that I didn't think of. Using a blind table to align the descriptions for example might work satisfactory.

ohdatboi:
Okay, so I made it nice and compact at the same time:
mypacks_edited

What I did is:
- Package name is a **bold** red link
- Version displayed as a **v1.0.0** or **~master**, because small badges are unreadable
- Description is just a text
- If there is an error, exclamation mark is displayed before package name (Should be replaced with an icon :warning:)

s-ludwig:
Thanks, that should work. I would drop the "v" prefix, though, what was the reason you added it? For narrow screens we could of course still break the description into a separate paragraph to give it more structure.

One thing that was quite helpful though is the actual list of errors, because it was possible to quickly tell whether the error is current or obsolete. But maybe we can only display the warning sign selectively in the future (or at least list the affected versions beneath the icon).

ohdatboi:
>  I would drop the "v" prefix, though, what was the reason you added it?

No idea.

> One thing that was quite helpful though is the actual list of errors

I removed it because it isn't obvious which package has an error - previous or next one. Maybe I can add border to make it look nice.

ohdatboi:
Looks ok IMO.

screenshot from 2017-08-30 13-16-36

s-ludwig:
It looks a bit crowded with all parts having the same font size and similar weight. The bold package names help a bit, but it's still much harder to scan through than the status-quo. Not sure about the case where the descriptions wraps to multiple lines, but a blind table layout with two columns (name/version + description) could also work as an intermediate solution (on narrow screens I think the initial version above is actually pretty good).

ohdatboi:
> blind table layout with two columns (name/version + description) could also work as an intermediate solution 

Tables are evil and that's why:
screencapture-code-dlang-org-my_packages-1504429387266

So, on small screens it's going to be horrible.

> on narrow screens I think the initial version above is actually pretty good

Indeed. But how do I use table on the big screens and list on the small ones? Looks like it'd require some JS, which is something I want to avoid.

s-ludwig:
Media queries plus `display: table(-row/-cell)` should work for this.

ohdatboi:
mypacks_list
mypacks_table

s-ludwig:
Nice, thanks! That's definitely a step forward over the status quo. We can tweak the error messages later, but this looks good to merge from my side.

PR #252, opened 2017-09-05
Compute statistics updates in a queue parallel to the general update queue.

s-ludwig:
While the update queue is mostly API performance dependent, the statistics updates are mostly DB query performance dependent. Computing them in sequence resulted in a serious decline in package update check performance. With this change, statistics updates should have almost no impact on the update speed. If they turn out to be slower than the updates, they will simply be updated as fast as possible in a round-robin fashion.


s-ludwig:
Yeah, update time went up from around 5-8 minutes to over half an hour recently. I didn't verify that it was caused by the stats code, but that was the only change in this area during that time.

Issue #58, opened 2014-05-23
Readme file isn't displayed unless named exactly README.md


ghost:
This means casing is important. Offending line:

``` D
try rep.readFile(reference.sha, Path("/README.md"), 
(scope input) { dbver.readme = input.readAllUTF8(); });
```

Note that (at least Github) allows all sorts of names for the readme: `readme.md`, `readme.markdown`, `README.md`, `README.markdown`, etc. I'm not sure whether all supported combinations are listed anywhere.

There's also `.rst` support on Github but I don't know if we have any D `.rst` parsers around for use with dub-registry.

PR #253, opened 2017-09-09
Fix anchor link in README.md when viewing package info

icy:
If the url is `#foobar` (an anchor link), the blob file should be `README.md`

### DIFF
--- a/source/dubregistry/web.d
+++ b/source/dubregistry/web.d
@@ -210,7 +210,7 @@ class DubRegistryWebFrontend {
 						// TODO: BitBucket + GitLab
 						case "github":
 							if (is_image) return format("https://github.com/%s/%s/raw/%s/%s", owner, project, gitVer, url);
-							else return format("https://github.com/%s/%s/blob/%s/%s", owner, project, gitVer, url);
+							else return format("https://github.com/%s/%s/blob/%s/%s", owner, project, gitVer, url.startsWith("#") ? "README.md" ~ url : url);
 					}
 				}
 


s-ludwig:
Wouldn't it be better to always return links starting with `#` as-is? That would enable anchor based in-page navigation within the package page on code.dlang.org instead of jumping to GitHub (plus it would also work for other hosters).

icy:
@s-ludwig That's a good idea and I think that will not confuse users.

However, the [Viber's markdown module](http://vibed.org/api/vibe.textfilter.markdown/) isn't _good_ when rendering Github README files. For example, when there are multiple headers with the same name Github knows how to generate different anchors while Viber doesn't

```
User input           Github output       Viber output

## Section 1         
### Introduction     #introduction       #introduction
### Conclusion       #conclusion         #conclusion

## Section 2
### Introduction     #introduction-1     #introduction
### Conclusion       #conclusion-1       #conclusion
```
If there is TOC (Table of Contents) in the README.md, links are almost broken with Viber output.

Using a constant `README.md` isn't good as in #58 .

s-ludwig:
> However, the Viber's markdown module isn't good when rendering Github README files. For example, when there are multiple headers with the same name Github knows how to generate different anchors while Viber doesn't

Good point. The question is also whether Bitbucket and Gitlab use the same logic to generate the headers, or if they differ. Makes sense to keep the current scheme for now then.

> Using a constant README.md isn't good as in #58 .

I've seen that GitHub has an API field for the readme file now, so this can be fixed pretty easily and in a more generic way.

icy:
> The question is also whether Bitbucket and Gitlab use the same logic to generate the headers, or if they differ. Makes sense to keep the current scheme for now then.

I don't know Bitbucket much. Gitlab and Github has almost the same way to generate anchors.

If you like to keep anchor links as-is, my patch will not be necessary. Both Gitlab and Github markdown have their own customized extensions [1][1] [2][2] so you will have more work to do :) Until the feature is implemented all anchor links in `REAMDE.md` are broken because they original filename is missing.

[1]: https://guides.github.com/features/mastering-markdown/ 
[2]: https://docs.gitlab.com/ee/user/markdown.html

s-ludwig:
No no, I meant the current scheme *within this pull request*. Imitating the anchor name generation of GitHub etc. can be done later, when the possible differences are well understood.

icy:
oh I see :) Thanks for your explanation.

PR #254, opened 2017-09-15
[trivial] Style disabled buttons gray and with no "hand" mouse cursor.

s-ludwig:
In particular makes the update section look like this instead of having a green button:
image

### DIFF
--- a/public/styles/common.css
+++ b/public/styles/common.css
@@ -412,6 +412,12 @@ div.packageInfo > dl > dd {
 	border-color: #4cae4c;
 }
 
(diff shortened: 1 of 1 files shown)


MartinNowak:
Are you selectively deploying changes? I already see this live, while the package rank feature is still missing.

s-ludwig:
Package rank has been there since quite a while (since shortly after the PR was pulled): https://code.dlang.org/?showrating

In this case I just pulled in the latest css, but usually updates are monolithic.

PR #258, opened 2017-10-24
Mention dub using pkg-config in the package file format documentations

kas-luthor:
The documentation didn't mention that pkg-config is used on Posix platforms to resolve linker flags for libraries specified in the "libs" field of the package config file


s-ludwig:
Thanks!

PR #209, opened 2017-05-28
Added Per Page option

WebFreak001:

PR #203, opened 2017-05-18
New layout of package page

ohdatboi:
I wanted to add some usefull little things like download statistics or stars, but realised that there is no place where I can put them.

So, I did this:
All package information moved to sidebar at right of the screen. On the small screens (<54em) it showed at the bottom of page. Package name, description, installation instructions and readme stays at left of the screen.
Also, sidebar is grey(gray?) and contains a lot of empty space, so it won't look heavy.

Anyway, it's not done yet
## TODO
- [x] On big screens looks ugly
- [x] Link to page of all versions in `Versions:` section
- [x] 'Versions:' section should also contain date of every version (Looks like a lot of work on backend)

PR #262, opened 2017-11-27
Reapply PR #209, which got accidentally reverted in #203 previously.

s-ludwig:
Added Per Page option

Limit should stay when changing categories

Added function to generate home url instead of copy pasting


s-ludwig:
That auto merge behavior with no required checks is really annoying...

MartinNowak:
Yes, most useless behavior https://github.com/dlang-bots/dlang-bot/issues/69.

PR #263, opened 2017-12-12
Change categories icons with font icons

WebFreak001:
New design:

Blue icons are packages tagged as application


WebFreak001:
bump @s-ludwig @wilzbach

MartinNowak:
Nice!
I'm hoping to replace the hardly usable/useful categories with tags at some point. Any idea how that would work for tags?

WebFreak001:
tags is a good idea, I would redesign it a bit to have some tags associated with icons and have the first tag with icon be the small icon of a package. It should show at least a few tags in the package listing

MartinNowak:
Yes, integration with a free-form search box is a bit tricky, but solvable.

PR #267, opened 2017-12-21
time to display score

MartinNowak:
- already affects search results
- we get no feedback/improvement ideas while score remains hidden
=> time to lift the curtains

### DIFF
--- a/source/dubregistry/dbcontroller.d
+++ b/source/dubregistry/dbcontroller.d
@@ -58,12 +58,16 @@ class DbController {
 		DbPackageStats stats;
 		m_packages.update(["stats": ["$exists": false]], ["$set": ["stats": stats]], UpdateFlags.multiUpdate);
 
-		float rating = 0;
-		m_packages.update(["stats.rating": ["$exists": false]], ["$set": ["stats.rating": rating]], UpdateFlags.multiUpdate);
+		// rename stats.rating -> stats.score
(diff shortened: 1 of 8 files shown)


s-ludwig:
We should give this metric a semantic name, the generic "rating" currently doesn't give a clue what it actually means. What about "relevancy"?

WebFreak001:
uh what unit is this even supposed to be, my most downloaded and one of my most useful packages has a rating of 1.8

s-ludwig:
It's in the range of zero to five (a 5-star rating was the initial idea).

MartinNowak:
Fairly straightforward, balancing a couple of features against the average.
https://github.com/dlang/dub-registry/blob/6e4724067693bab226894435fe4538beca289957/source/dubregistry/registry.d#L575-L617

> uh what unit is this even supposed to be, my most downloaded and one of my most useful packages has a rating of 1.8

Which one?

WebFreak001:
@MartinNowak https://github.com/WebFreak001/FSWatch

MartinNowak:
> @MartinNowak https://github.com/WebFreak001/FSWatch

@WebFreak001 while 10K downloads is quite a lot, there were only 11 downloads in the past month (30 days) which are relevant for the rating.

MartinNowak:
Ready from my side @s-ludwig.
Starting with the migrated data but an old mirror seems to work (just tested), as the score is recomputed on startup. Please still let us know before you're deploying this (e.g. via Slack). FYI @wilzbach

s-ludwig:
Looks good now!

PR #272, opened 2018-01-27
Change the license to Boost

wilzbach:
From https://github.com/dlang/dub/issues/1349

> Thanks for the info. I am not sure whether these commercial companies will look into the source code because dub registry is under GPL license. Maybe you can check whether dub registry could change the license?

CC @s-ludwig @MartinNowak @ohdatboi @ColdenCullen @WebFreak001 @brad-anderson @mdondorff @Kripth @JakobOvrum - are you all okay with this?

--- edit:

The GPL license was added to the `dub.sdl` in this commit: https://github.com/dlang/dub-registry/commit/cc08b3509c3107f1d563b69db3d33c0695b35475 and the LICENSE.txt in this commit https://github.com/dlang/dub-registry/commit/70499db0f9e8eed7cabb36e03d04ccb44f53e105


ColdenCullen:
This is fine by me! I personally am a huge fan of the Boost license.

s-ludwig:
Still missing @MartinNowak, @brad-anderson, @mdondorff, @Kripth, and @JakobOvrum.

brad-anderson:
I hereby authorize my typo fixes to be relicensed under the Boost license 😜

mdondorff:
I authorize, thanks!

Kripth:
I do authorise my changes to be relicensed.

s-ludwig:
@MartinNowak can we assume that you agree, too? Looking at the log, the changes by @JakobOvrum look like they are not subject to copyright issues, so it should be safe to continue in that case.

MartinNowak:
Fine with me.

WalterBright:
Thank you everyone! This is good news indeed.

PR #274, opened 2018-01-31
Add <package>.dub.pm shorturls to the package overview

wilzbach:
Preview:

image

(FWIW the dub registry doesn't work well with mongo 3.6 - I had to disable the stats aggregation locally for now.)

### DIFF
--- a/views/view_package.dt
+++ b/views/view_package.dt
@@ -177,6 +177,10 @@ block body
 			dt Score:
 			dd#score(title="#{formatPackageStats(stats)}")!= formatScore(req.rootDir, stats.score)
 
+			dt Short URL:
+				dd
+					a(href="http://#{packageName}.dub.pm", target="_blank") #{packageName}.dub.pm
+
 	script(type="application/javascript", src="/scripts/clipboard.min.js")
 	:javascript
 		new Clipboard('.btn-clipboard');


views/view_package.dt
```diff
@@ -177,6 +177,10 @@ block body
 			dt Score:
 			dd#score(title="#{formatPackageStats(stats)}")!= formatScore(req.rootDir, stats.score)
 
+			dt Short URL:
+				dd
+					a(href="http://#{packageName}.dub.pm", target="_blank") #{packageName}.dub.pm
```
wilzbach: Both http://mir.dub.pm and http://dub.pm/p/mir work - I thought sub-domains are easier.

In case someone worries about SSL - see https://letsencrypt.org/2017/07/06/wildcard-certificates-coming-jan-2018.html (tl;dr; this should be possible soon)

wilzbach:
Ping @s-ludwig - are you ok with this or do you have a better idea?

s-ludwig:
I haven't really made my mind up about this. Are we absolutely sure that we won't need any of the sub domains for special purposes? If so, it looks rather unclean to me to mix that with user generated domains. But maybe putting the homepage on https://dub.pm/ is everything that's really still missing from dub.pm anyway, in which case I have no objections.

wilzbach:
> Are we absolutely sure that we won't need any of the sub domains for special purposes?

Well the only thing that comes to my mind are mirrors, but `code-mirrorX.dlang.io` has worked fine so far.
If someone adds a new mirror, having it at a separate DNS registry might be even better for additional redundancy.

Other potential usages for special sub domains
- docs.dub.pm (i.e. what dpldocs.info is currently doing)
- cdn.dub.pm (caches the zip downloads)

But with is going to happen soon and we couldn't use `dlang.io` (or `dub.dlang.io`) for.

> But maybe putting the homepage on https://dub.pm/ is everything that's really still missing from dub.pm anyway, in which case I have no objections.

BTW the newly added netlify:

https://dub-registry.netlify.com/getting_started.html

I just changed the DNS of the root-level + `www` domain to point to netlify, so the following should work in a few hours as the HTML we generate for the previews is auto-deployed to them:

https://dub.pm/getting_started.html

(though we probably still want to move all these help pages into a different project or at least sub-project, s.t. they don't slow-down the default build-time when generating the diet templates)

wilzbach:
> btw we should block package names such as www

Yep, that's hard-coded already (in the DNS registar and now points to netlify)

> Also how is the letsencrypt coming along?

Apparently it's live since two weeks: 

https://letsencrypt.org/2017/07/06/wildcard-certificates-coming-jan-2018.html
https://community.letsencrypt.org/t/acmev2-and-wildcard-launch-delay/53654
https://community.letsencrypt.org/t/acme-v2-and-wildcard-certificate-support-is-live/55579

Will check if it works and I can enable it :)

wilzbach:
Looks like I was successful: https://mir.dub.pm

PR #277, opened 2018-02-09
optionally minimize requested package info fields

MartinNowak:
- strips readmes for now
- also be explicit internally


MartinNowak:
Did it in #278, let's close this sidekiq PR.

PR #278, opened 2018-02-09
add API to get info for package and dependencies

MartinNowak:
Allows us to use a single request in dub.

also see #277 for the minimize commit


source/dubregistry/api.d
```diff
@@ -145,16 +148,24 @@ override {
...
+		return m_registry.getPackageVersionInfo(rootOf(name), ver, minimize)
 			.check!(r => r.type != Json.Type.null_)(HTTPStatus.notFound, "Package/Version not found");
 	}
+
+	Json[string] getPackageInfoWithDependencies(string name, bool minimize = false) {
+		import std.array : assocArray;
+
+		auto aa = m_registry.getPackageInfoWithDependencies(rootOf(name), false, minimize)
+			.check!(r => r !is null)(HTTPStatus.notFound, "Package/Version not found");
+		return aa.byKeyValue.map!((p) => tuple(p.key, p.value.info)).assocArray;
```
wilzbach: `assocArray` won't make the GC leaks easier ... it's a shame that streaming Json would require a serious refactoring :/

source/dubregistry/registry.d
```diff
@@ -248,6 +250,44 @@ class DubRegistry {
 		return ret;
 	}
 
+	PackageInfo[string] getPackageInfoWithDependencies(string packname, bool include_errors, bool minimize)
+	{
+		PackageInfo[string] ret;
+		getPackageInfoWithDependencies(packname, include_errors, minimize, ret);
+		logDebug("%s", ret.keys);
```
wilzbach: Maybe add a tag to grep for it in the log?
e.g. `getPackageInfoWithDependencies[keys]: `

source/dubregistry/registry.d
```diff
@@ -418,7 +458,7 @@ class DubRegistry {
 		string[] errors;
 
 		PackageInfo pack;
-		try pack = getPackageInfo(packname);
+		try pack = getPackageInfo(packname, false, true);
```
wilzbach: Vibe.d should really support `Flag` out of the box :O

source/dubregistry/registry.d
```diff
@@ -248,6 +250,44 @@ class DubRegistry {
...
+		{
+			foreach (dep, _; v.info["dependencies"].opt!(Json[string]))
+			{
+				auto depname = rootOf(dep);
+				if (depname !in ret)
+					getPackageInfoWithDependencies(depname, include_errors, minimize, ret);
+			}
+			foreach (cfg; v.info["configurations"].opt!(Json[]))
+			{
+				foreach (dep, _; v.info["dependencies"].opt!(Json[string]))
```
wilzbach: This looks wrong. Shouldn't it be `cfg.info`?

MartinNowak: Yes, fixed

source/dubregistry/api.d
MartinNowak: I think the leak has a very defined location, likely a buffer in the mongodb driver.
Collecting those return values should be rather trivial. Also it's a thin copy, the JSON values remain the same.

MartinNowak:
BTW, dub already transparently supports transfer encoding (compression). This was enabled in https://github.com/dlang/dub/pull/1247.

MartinNowak:
Ah wait, this doesn't fully work with multiple root dependencies.

MartinNowak:
I did a fairly aggressive minimization reducing ~900KiB (80KiB compressed) for dub and dependencies down to ~200KiB (5.4KiB compressed).
Still blocked until the dub side is finished as well.

source/app.d
```diff
@@ -109,6 +109,7 @@ shared static this()
 	settings.bindAddresses = ["127.0.0.1"];
 	settings.port = 8005;
 	settings.sessionStore = new MemorySessionStore;
+	settings.useCompressionIfPossible = true;
```
s-ludwig: There has been a particular fix in the ZLIB stream implementation that may have eliminated the main culprit, but using compression in has led to occasional transfer errors in the past. In order to minimize the risk we could just selectively set the compression header in the `getInfo` end point. We could also just observe whether it actually still has any issues, though, because knowing that would also be quite interesting.

It needs to be said that this setting will generally not have an effect on its own, though, because the final behavior depends on the reverse proxy facing the internet.

source/dubregistry/registry.d
```diff
@@ -570,6 +655,15 @@ struct PackageInfo {
 	Json info; /// JSON package information, as reported to the client
 }
 
+/// flags to customize getPackageInfo* methods
+enum PackageInfoFlags
+{
+	none,
+	include_dependencies = 1 << 0, /// include package info of dependencies
```
s-ludwig: Naming convention: `includeDependencies`/`includeErrors`

source/app.d
wilzbach: > It needs to be said that this setting will generally not have an effect on its own, though, because the final behavior depends on the reverse proxy facing the internet.

Yeah, shouldn't this be handled entirely by the reverse proxy?

source/dubregistry/api.d
```diff
@@ -145,16 +149,30 @@ override {
...
+	Json[string] getInfos(string[] packages, bool include_dependencies = false, bool minimize = false)
+	{
+		import std.array : assocArray;
+
+		auto flags = minimize ? PackageInfoFlags.minimize : PackageInfoFlags.none;
+		if (include_dependencies)
+			flags |= PackageInfoFlags.include_dependencies;
+		return m_registry.getPackageInfos(packages, flags)
+			.check!(r => r !is null)(HTTPStatus.notFound, "None of the packages were found")
+			.byKeyValue.map!(p => tuple(p.key, p.value.info)).assocArray;
```
wilzbach: FYI: https://github.com/dlang/phobos/pull/6155 (makes this easier )

source/dubregistry/registry.d
```diff
@@ -248,6 +277,62 @@ class DubRegistry {
 		return ret;
 	}
 
+	PackageInfo[string] getPackageInfos(string[] packnames, PackageInfoFlags flags)
+	{
+		import dub.recipe.packagerecipe : getBasePackageName;
+
+		PackageInfo[string] infos;
+		void[0][string] visited;
```
wilzbach: Walter wants to deprecate this. 

https://github.com/dlang/dmd/pull/7986

Why not `bool`?

source/dubregistry/registry.d
wilzbach: Fixed this (only for this flag, not for the web interface)

source/dubregistry/registry.d
```diff
@@ -248,6 +277,62 @@ class DubRegistry {
...
+		import std.algorithm.searching : find;
+
+		if (packname in visited)
+			return;
+		visited[packname] = typeof(visited[packname]).init;
+
+		auto basepkg = getBasePackageName(packname);
+		auto p = basepkg in infos;
+		if (p is null)
+			p = &(infos[basepkg] = getPackageInfo(basepkg, flags));
```
wilzbach: `p` is never used below - so this could be:

```d
if (basepkg in infos)
    infos[basepkg] = ...
```

source/dubregistry/registry.d
```diff
@@ -248,6 +277,62 @@ class DubRegistry {
...
+				foreach (dep, _; cfg["dependencies"].opt!(Json[string]))
+					getPackageInfos(dep, flags, infos, visited);
+		}
+
+	Louter: foreach (v; p.versions)
+		{
+			auto info = v.info;
+			foreach (subpkg; packname.splitter(":").dropOne)
+			{
+				auto sp = info["subPackages"].opt!(Json[]).find!(sp => sp["name"] == subpkg);
```
wilzbach: `if info["subPackages"][].find` would work too

source/dubregistry/api.d
```diff
@@ -147,16 +151,30 @@ override {
...
+	Json[string] getInfos(string[] packages, bool include_dependencies = false, bool minimize = false)
+	{
+		import std.array : assocArray;
+
+		auto flags = minimize ? PackageInfoFlags.minimize : PackageInfoFlags.none;
+		if (include_dependencies)
+			flags |= PackageInfoFlags.includeDependencies;
+		return m_registry.getPackageInfos(packages, flags)
+			.check!(r => r !is null)(HTTPStatus.notFound, "None of the packages were found")
+			.byKeyValue.map!(p => tuple(p.key, p.value.info)).assocArray;
```
wilzbach: FYI: https://github.com/dlang/phobos/pull/6155 (makes this easier )

wilzbach:
Also rebased this to master.

source/dubregistry/registry.d
```diff
@@ -275,6 +304,64 @@ class DubRegistry {
 		return ret;
 	}
 
+	PackageInfo[string] getPackageInfosRecursive(string[] packnames, PackageInfoFlags flags)
```
wilzbach: https://github.com/dlang/dub-registry/commit/ca5d2670533d2055fc784157683af01efafaa288#diff-31fe65cc39b31f74454f75b09e88a761 has been added in the meantime and to avoid name conflicts I renamed this method.
It might make sense to unify these, but this one here returns an eager array whereas the other returns a lazy range.

source/dubregistry/registry.d
```diff
@@ -275,6 +304,64 @@ class DubRegistry {
...
+		import std.range : dropOne;
+		import std.algorithm.searching : find;
+
+		if (packname in visited)
+			return;
+		visited[packname] = typeof(visited[packname]).init;
+
+		auto basepkg = () @trusted {
+			return getBasePackageName(packname);
+		}();
```
wilzbach: Apparently this regressed from @safe to @system from dub 1.3 to dub 1.7
Fix: https://github.com/dlang/dub/pull/1432

s-ludwig:
Anything holding this back apart from a rebase (removing the `@trusted` wrapper)?

wilzbach:
> Anything holding this back apart from a rebase (removing the @trusted wrapper)?

I don't think so. Just removed the @trusted wrapper ;-)

wilzbach:
Resolved the merge conflicts. Any chance we can get this deployed in the near future?
It would really help a lot with dub's performance...

wilzbach:
Finally got the Heroku GitHub PR preview working: https://dub-registry-staging-pr-278.herokuapp.com
Though without https://github.com/dlang/dub-registry/pull/361, we can't register packages.

wilzbach:
One can actually nicely test this with the new Heroku PR preview pipeline.

An example:

https://dub-registry-staging-pr-278.herokuapp.com/api/packages/rivald/info
https://dub-registry-staging-pr-278.herokuapp.com/api/packages/rivald/info?minimize=true

wilzbach:
> Anything holding this back apart from a rebase

Nope. Did a final rebase and squashed some of my commits as the `@trusted` wrapper is now a thing of the past,

wilzbach:
Ping @s-ludwig - any chance we could get this deployed? It would allows us to enable this in dub and ship this with 2.082 (dub 1.11)

PR #286, opened 2018-02-26
Allow custom package logos & add OpenGraph

WebFreak001:
This is a split part of #265, just only for package logos

This commit goes slightly different on the caching name due to the comments, but still keeps all image files. This time the user specifies the revision of the icon and gets the default package if it is not found. If the user does not specify one they get the current logo. I decided to go for this approach due to archiving or embedding purposes, especially on google etc. where an old image of the repository might be wanted.

---

---

@MartinNowak @wilzbach


source/dubregistry/web.d
```diff
@@ -139,6 +140,42 @@ class DubRegistryWebFrontend {
...
+			if (logo.length)
+			{
+				res.redirect(logo);
+				return;
+			}
+		}
+
+		string logo = _packname ~ "@" ~ rev;
+
+		bool[logoFormats.length] exists;
```
wilzbach: `exists` is a bad name here. `imageFiles`, `images`, ...?

source/dubregistry/web.d
```diff
@@ -139,6 +140,42 @@ class DubRegistryWebFrontend {
...
+		NativePath[logoFormats.length] paths;
+		foreach (i, format; logoFormats)
```
wilzbach: `static foreach`

source/dubregistry/web.d
```diff
@@ -139,6 +140,42 @@ class DubRegistryWebFrontend {
...
+		{
+			paths[i] = NativePath(logoOutputFolder) ~ NativePath.Segment(logo ~ format);
+			exists[i] = existsFile(paths[i]);
+		}
+
+		auto settings = new HTTPFileServerSettings();
+		settings.maxAge = 365.days;
+
+		if (exists[0])
+			sendFile(req, res, paths[0], settings);
```
wilzbach: So we just did a lot of file checks, but send the first one anyhow?

source/dubregistry/web.d
```diff
@@ -595,6 +632,29 @@ class DubRegistryFullWebFrontend : DubRegistryWebFrontend {
 		redirect("/my_packages/"~_packname);
 	}
 
+	@auth @path("/my_packages/:packname/set_logo")
+	void postSetLogo(scope HTTPServerRequest request, string _packname, User _user)
+	{
+		enforceUserPackage(_user, _packname);
```
wilzbach: That would be an excellent use-case for either an `auth.Owner` role or a `@before` tag, but that's over the scope of this PR.

uploads/logos/.gitignore
```diff
@@ -0,0 +1 @@
+*.png
```
wilzbach: Did you already made an argument for storing this locally an not in mongo?
I know that mongo isn't ideal, but it probably makes backups a lot easier (and developing with a mirror).

views/view_package.dt
```diff
@@ -10,6 +10,25 @@ block title
...
+	meta(property="og:title", content="Package #{packageName.split(':')[0]} on DUB")
+	meta(property="og:image", content=packageInfo["logo"].opt("#{req.rootDir}public/images/default-logo.png"))
+	meta(property="og:description", content=versionInfo["description"].opt!string)
+	meta(property="og:site_name", content="DUB Package Registry")
+	meta(property="og:locale", content="en_US")
+	meta(property="article:author", content=user.fullName.length ? user.fullName : user.name)
+
+	// twitter
+	meta(name="twitter:card", content="summary")
+	//- TODO: register dub twitter account
```
wilzbach: I'm not sure whether it's worthwhile to create a dub registry account
BTW tweeting about every new package is probably not very helpful either.
[dlangbot](https://twitter.com/dlangbot) doesn't have many followers.

views/view_package.dt
```diff
@@ -10,6 +10,25 @@ block title
 	- import userman.controller;
 	- auto title = "Package " ~ packageName ~ " version " ~ versionInfo["version"].get!string;
 
+	// facebook & co.
+	meta(property="og:url", content="http#{req.tls ? `s` : ``}://#{req.host}/packages/#{packageName.split(':')[0]}")
```
wilzbach: how about wrapping the host things into a functions? `#{req.registryURL}`

views/view_package.dt
```diff
@@ -10,6 +10,25 @@ block title
...
+	meta(property="og:type", content="article")
+	meta(property="og:title", content="Package #{packageName.split(':')[0]} on DUB")
```
wilzbach: `splitter` and do we really have to use split here?

views/view_package.dt
```diff
@@ -10,6 +10,25 @@ block title
...
+	meta(property="og:site_name", content="DUB Package Registry")
+	meta(property="og:locale", content="en_US")
+	meta(property="article:author", content=user.fullName.length ? user.fullName : user.name)
+
+	// twitter
+	meta(name="twitter:card", content="summary")
+	//- TODO: register dub twitter account
+	meta(name="twitter:site", content="@D_Programming")
+	meta(name="twitter:url", content="http#{req.tls ? `s` : ``}://#{req.host}/packages/#{packageName.split(':')[0]}")
+	meta(name="twitter:title", content="Package #{packageName.split(':')[0]} on DUB")
```
wilzbach: `splitter` and do we really have to use split here?

views/view_package.dt
```diff
@@ -21,6 +40,11 @@ block body
 		- if (req.session && user.id == User.ID.fromString(req.session.get!string("userID")))
 			a(href="#{req.rootDir}my_packages/#{packageName.split(':')[0]}") Manage this package
 
+		- auto logo = packageInfo["logo"].opt("");
+
+		- if (logo.length)
+			img.packageLogo(src=logo, alt="#{packageName.split(':')[0]} logo")
```
wilzbach: `splitter` and do we really have to use split here?

design/package-logo.svg
```diff
@@ -0,0 +1,374 @@
...
+     inkscape:label="Layer 2"
+     sodipodi:insensitive="true"
+     style="display:none">
+    <image
+       transform="translate(0,-229.26665)"
+       y="237.1985"
+       x="10.739017"
+       id="image5884"
+       xlink:href="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAg0AAAIJCAIAAACp65nVAAAAA3NCSVQICAjb4U/gAAAgAElEQVR4
+nO3dyW8beX7//6rivkmkREryItNLe2u7DTc83ZPpZIJJgEGOQc7Jf5BzECT/RoCcghxyCJIAOc0E
```
wilzbach: Why do you use inline png here?

source/dubregistry/internal/utils.d
```diff
@@ -13,3 +23,56 @@ string black(string url)
...
+static immutable string logoOutputFolder = "uploads/logos";
+
+static immutable string[] logoFormats = [".png"];
+
+/// 
+/// Params:
+///   file = the file to convert
+///   name = the logo name to put in uploads/logos
+///   deleteExisting = if false, throw an exception if the files already exist
+///   deleteFinish = if true, delete the input file after at least one successful conversion
```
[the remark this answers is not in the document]
s-ludwig: Yeah, unfortnately it wasn't possible so far to persuade Walter into fixing the semantics for this. What this will do is to create one paragraph per line instead of a single one.

source/dubregistry/web.d
```diff
@@ -139,6 +140,42 @@ class DubRegistryWebFrontend {
...
 
+	@path("/packages/:packname/logo")
+	void getPackageLogo(HTTPServerRequest req, HTTPServerResponse res, string _packname)
+	{
+		string rev = req.queryString;
+		if (!rev.length)
+		{
+			// get latest revision of file
+			auto logo = m_registry.getPackageLogo(_packname);
+			if (logo == req.requestPath.toString)
```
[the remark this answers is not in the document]
s-ludwig: `InetPath.toString()` doesn't allocate anymore, so `toString` in general should be fine to use. We could add a new property `.stringRepresentation` or similar to make this more clear, though.

s-ludwig:
Agree with the default logo. Reducing opacity and/or adding a checker pattern or similar could be used to signal that it's a default logo, too.

source/dubregistry/internal/utils.d
```diff
@@ -13,3 +23,56 @@ string black(string url)
...
+		string base = buildPath(logoOutputFolder, name);
+		string pngOutput = base ~ ".png";
+		auto png = spawnProcess(["convert", file.toNativeString, "-resize", "512x512>", pngOutput]);
+
+		if (png.wait == 0)
+			success[0] = true;
+	}
+	catch (Exception e)
+	{
+		(() @trusted => send(owner, e.toString()))();
```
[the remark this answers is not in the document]
s-ludwig: I think it should, but the vibe.d's earlier re-implementation before the compatibility additions, didn't have that feature, so all examples are still explicitly passing the owner (opened https://github.com/vibe-d/vibe.d/issues/2091).

s-ludwig:
BTW, a relatively cheap way to instantly get some non-default icons would be to use the organization avatar on GitHub (not if the owner is a user obviously, though) and the project avatar for GitLab and Bitbucket. Could be a new field in `dubregistry.repositories.repository.RepositoryInfo`.

source/dubregistry/internal/utils.d
```diff
@@ -13,3 +23,56 @@ string black(string url)
...
+{
+	try
+	{
+		bool[logoFormats.length] success;
+		scope (success)
+			(() @trusted => send(owner, success))();
+
+		string base = buildPath(logoOutputFolder, name);
+		string pngOutput = base ~ ".png";
+		auto png = spawnProcess(["convert", file.toNativeString, "-resize", "512x512>", pngOutput]);
```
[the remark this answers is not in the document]
WebFreak001: the > makes sure aspect ratio is kept, it's not a pipe and spawnProcess couldn't pipe anyway

source/dubregistry/web.d
WebFreak001: uh relict from other PR where we had more than 1 file check, we actually just do one file check at the above too, maybe support for jpg or other formats will be added eventually and this is the place to put it

design/package-logo.svg
WebFreak001: uhh, dunno what inkscape did there. I had the D logo in there once but it didn't look so good, maybe it still kept it

public/styles/common.css
```diff
@@ -276,6 +276,30 @@ button {
...
+	max-width: 150px;
+	max-height: 150px;
+	margin: 0 0 1em 1em;
+}
+
+@media only screen and (max-width: 480px) {
+	img.packageLogo {
+		display: block;
+		float: none;
+		max-width: 100%;
```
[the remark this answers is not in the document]
WebFreak001: made it so it isn't higher than 220px now, it can still be a wide logo but square logo will be smaller now. It does not stretch (at least not in chrome)

WebFreak001:
@s-ludwig made the default logo gray

WebFreak001:
I think getting a default logo on registration time or something like that is scope of another PR.

WebFreak001:
mergable now?

s-ludwig:
I would second the question of whether to store logos as files or in the database. For the mentioned reasons I'd lean towards database, it also won't work with the current setup, because the registry runs as an unprivileged user without write access to its own directory.

BTW, does the `convert` invocation always scale to 512, or will it only do downscaling?

Apart from that I'd say everything is ready.

WebFreak001:
yeah ok I will check how well that is doable.

`convert` with `>` only downscales

views/view_package.dt
```diff
@@ -10,6 +10,28 @@ block title
...
+
+	// facebook & co.
+	meta(property="og:url", content="")
+	meta(property="og:type", content="article")
+	meta(property="og:title", content="Package #{normalizedPackageName} on DUB")
+	meta(property="og:image", content=packageInfo["logo"].opt("#{req.rootDir}public/images/default-logo.png"))
+	meta(property="og:description", content=versionInfo["description"].opt!string)
+	meta(property="og:site_name", content="DUB Package Registry")
+	meta(property="og:locale", content="en_US")
+	meta(property="article:author", content=user.fullName.length ? user.fullName : user.name)
```
MartinNowak: Nice :)

views/view_package.dt
```diff
@@ -10,6 +10,28 @@ block title
 	- import userman.controller;
 	- auto title = "Package " ~ packageName ~ " version " ~ versionInfo["version"].get!string;
 
+	- auto normalizedPackageName = packageName.splitter(':').front;
```
MartinNowak: `findSplit(':')[0]`

s-ludwig:
> Just scrape og:image from the repo sites initially to bootstrap the logos, then we don't need a repetitive default logo.

Didn't realize that they augmented the repo size with that! Nice.

wilzbach:
FYI: the default github avatars use a "technology" that's called identicon. You can even skip the query to github and create such icon on-demand, e.g.

https://identicon-api.herokuapp.com/dub-registry/200?format=svg

Or a newer service: https://jdenticon.com/#icon-dub-registry

See also:
- http://identicon.net/
- https://en.wikipedia.org/wiki/Identicon

views/view_package.dt
wilzbach: What's the advantage of findSplit?

MartinNowak:
> FYI: the default github avatars use a "technology" that's called identicon. You can even skip the query to github and create such icon on-demand, e.g.

We would just fetch the logo once for each package initially and on registration of new packages, really trivial and saves a manual step in case the organization/user logo is already the right choice and continues to work across github/gitlabs/bitbucket's feature set (e.g. per-repo logos https://gitlab.com/gitlab-org/gitter).

WebFreak001:
logos are now stored inside mongo as bson binary objects, however caching got removed due to this as sendFIle isn't used anymore.

public/styles/common.css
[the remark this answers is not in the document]
wilzbach: Shouldn't we require logos to be square and convert them otherwise? This usually makes such design things a lot easier.

source/dubregistry/dbcontroller.d
```diff
@@ -177,6 +177,35 @@ class DbController {
...
+			m_packages.update(["name": packname], ["$set": [
+				"logo": BsonBinData(BsonBinData.Type.generic, png),
+				"logoHash": BsonBinData(BsonBinData.Type.md5, cast(bdata_t)md5Of(png)[].idup)
+			]]);
+		}
+		else
+			m_packages.update(["name": packname], ["$unset": ["logo": 1, "logoHash": 1]]);
+	}
+
+	bdata_t getPackageLogo(string packname, out bdata_t rev)
```
wilzbach: This a question of design/personal preferences, but I typically return tuples in such cases. Hence out of interest, why did you go with `out`?

source/dubregistry/internal/utils.d
```diff
@@ -13,3 +26,42 @@ string black(string url)
...
+	}
+
+	auto a = appender!(ubyte[])();
+
+	(() @trusted {
+		foreach (chunk; png.stdout.byChunk(4096))
+			a.put(chunk);
+	})();
+
+	return cast(bdata_t)a.data.idup;
```
wilzbach: Appender uses the GC to allocate plus it's local, so you can use `assumeUnique` and avoid the idup copy

source/dubregistry/registry.d
```diff
@@ -7,14 +7,17 @@ module dubregistry.registry;
...
 import dubregistry.cache : FileNotFoundException;
 import dubregistry.dbcontroller;
+import dubregistry.internal.utils;
 import dubregistry.internal.workqueue;
 import dubregistry.repositories.repository;
 
 import dub.semver;
 import dub.package_ : packageInfoFilenames;
-import std.algorithm : canFind, countUntil, filter, map, sort, swap;
+import std.algorithm : any, canFind, countUntil, filter, map, sort, swap;
```
wilzbach: Any isn't used anymore

source/dubregistry/web.d
```diff
@@ -139,6 +140,33 @@ class DubRegistryWebFrontend {
...
+
+		if (logo.length) {
+			// TODO: add caching, etag (using rev), content-control, etc.
+			res.writeBody(logo, "image/png");
+		} else {
+			bool acceptsSVG;
+			// make sure requester actually supports svg, if a custom IDE or tool for fetching images is used it should only get png if it doesn't support svg.
+			// if requester is an IDE sending Accept: */*, but not accepting SVG it's their own fault.
+			foreach (accept; req.headers.get("Accept", "").splitter(",")) {
+				if (accept.startsWith("image/*", "image/svg", "*/*", "*/svg")) {
```
wilzbach: Isn't this `among`?

source/dubregistry/web.d
```diff
@@ -595,6 +623,30 @@ class DubRegistryFullWebFrontend : DubRegistryWebFrontend {
 		redirect("/my_packages/"~_packname);
 	}
 
+	@auth @path("/my_packages/:packname/set_logo")
+	void postSetLogo(scope HTTPServerRequest request, string _packname, User _user)
+	{
+		enforceUserPackage(_user, _packname);
+		const(FilePart) logo = request.files.get("logo");
```
wilzbach: FYI the braces aren't needed here. In fact, just `const` should work too.

source/dubregistry/web.d
```diff
@@ -595,6 +623,30 @@ class DubRegistryFullWebFrontend : DubRegistryWebFrontend {
...
+		enforceBadRequest(logo != FilePart.init);
+		auto renamed = NativePath.fromString(logo.tempPath.toString ~ logo.filename.name.extension);
```
wilzbach: `NativePath.fromString` should ideally accept a range...

source/dubregistry/registry.d
```diff
@@ -290,6 +294,25 @@ class DubRegistry {
...
+	void setPackageLogo(string pack_name, NativePath path)
+	{
+		auto png = generateLogo(path);
+		if (png.length)
+			m_db.setPackageLogo(pack_name, png);
+		else
+			throw new Exception("Failed to generate logo");
+	}
+
+	void unsetPackageLogo(string pack_name)
```
wilzbach: FWIW this is typically called delete (though I don't want to get into this nitpicking)

source/dubregistry/web.d
```diff
@@ -595,6 +623,30 @@ class DubRegistryFullWebFrontend : DubRegistryWebFrontend {
 		redirect("/my_packages/"~_packname);
 	}
 
+	@auth @path("/my_packages/:packname/set_logo")
```
wilzbach: FWIW this is usually mapped on `/logo` and only the http method is used to distinguish between get, post and delete.

source/dubregistry/internal/utils.d
s-ludwig: Or change to `appender!(immutable(ubyte)[])`

source/dubregistry/web.d
s-ludwig: Definitely wouldn't hurt from a convenience POV, but it stores a `string` inside anyway, so that it couldn't eliminate the allocation.

source/dubregistry/web.d
WebFreak001: `image/*; q=0.8`

source/dubregistry/web.d
WebFreak001: well all other endpoints are like this too

source/dubregistry/dbcontroller.d
WebFreak001: I go `struct` or `out`, but never tuple because it requires me to import something rather large. Selective imports don't help with compilation time in most of phobos.

source/dubregistry/web.d
WebFreak001: eventually gonna make a PR to vibe.d for proper Accept parsing

source/dubregistry/web.d
WebFreak001: or maybe some rename filename/last segment? (or any segment even)

public/styles/common.css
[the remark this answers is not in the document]
WebFreak001: hm I'm actually a fan of supporting flat logos such as banners, isn't that much more work with html and css

source/dubregistry/internal/utils.d
wilzbach: well, in this case you could simply use `execute` - this would save a few lines?
(I will look into making modifying `execute`)

WebFreak001: execute converts to a string, but the std output gets a binary png file.

source/dubregistry/dbcontroller.d
wilzbach: `std.typecons` is already imported in other modules (and even in vibe.d), so there's no extra import cost.
Also FWIW we reduced the total import time of all Phobos modules to < 0.5s with 2.079. On my machine it's even less than 0.2s now

See: https://dlang.org/changelog/2.079.0.html#std-experimental-all

source/dubregistry/web.d
wilzbach: > so that it couldn't eliminate the allocation.

Hmm, well I guess it could be templatized (would be nice with the hopefully soon upcoming RCString), but it's a different story entirely.
Anyhow I opened an issue, s.t. it's not totally forgotten (https://github.com/vibe-d/vibe.d/issues/2119).

source/dubregistry/internal/utils.d
wilzbach: > I don't want to risk an eventual std.utf.validate getting in there.

FTR with `cast(immutable(ubyte)[]) execute(["convert", ...])` `validate` isn't called.
Appender uses `ElementEncodingType`, not `ElementType` to avoid auto-decoding.

s-ludwig: Just for completeness, there is `vibe.stream.operations.readAll` to read all of a stream into a buffer.

WebFreak001: does it work for std.stdio.File like you get when spawning a process?

s-ludwig: No, would have to be wrapped inside a `StdFileStream` for that, but that is rather ineffecient currently.

source/dubregistry/web.d
s-ludwig: Yeah, let's keep that for another day. The current scheme is just used to avoid HTTP method override hacks to account for the form based nature of the interface.

source/dubregistry/dbcontroller.d
```diff
@@ -177,6 +177,35 @@ class DbController {
...
+
+		if (png.length)
+		{
+			m_packages.update(["name": packname], ["$set": [
+				"logo": BsonBinData(BsonBinData.Type.generic, png),
+				"logoHash": BsonBinData(BsonBinData.Type.md5, cast(bdata_t)md5Of(png)[].idup)
+			]]);
+		}
+		else
+			m_packages.update(["name": packname], ["$unset": ["logo": 1, "logoHash": 1]]);
```
wilzbach: It's quite untypical to use `1` as null value.

source/dubregistry/dbcontroller.d
```diff
@@ -177,6 +177,35 @@ class DbController {
...
+		auto bpack = m_packages.findOne(["name": packname]);
+		if (bpack.isNull)
+			return null;
+		auto logo = bpack.tryIndex("logo");
+		if (logo.isNull)
+			return null;
+		auto hash = bpack.tryIndex("logoHash");
+		if (!hash.isNull)
+			rev = hash.get.get!BsonBinData.rawData;
+		return logo.get.get!BsonBinData.rawData;
```
wilzbach: We should add [safeDeref](https://forum.dlang.org/post/fshlmahxfaeqtwjbjouz@forum.dlang.org) to Vibe.d, it could then be sth. like:

```
rev = bpack.try.logoHash.rawData.get!BsonBinData;
return bpack.try.logo.rawData.get!BsonBinData;
```

views/my_packages.package.dt
```diff
@@ -19,6 +19,18 @@ block body
...
 				br
 
+	h2 General
+
+	.inputForm
+		form(method="POST", enctype="multipart/form-data", action="#{req.rootDir}my_packages/#{packageName}/set_logo")
+			h3 Logo
+			- string logoPath = "/packages/" ~ packageName ~ "/logo";
+			img.packageLogo(src=logoPath)
+			input(name="logo", type="file", accept="image/*", onchange="updateLogoPreview(this)")
```
wilzbach: Maybe `accept="image/x-png,image/jpeg"` or `accept=".png, .jpg, .jpeg" ` for now?

source/dubregistry/web.d
```diff
@@ -595,6 +623,30 @@ class DubRegistryFullWebFrontend : DubRegistryWebFrontend {
...
+	void postSetLogo(scope HTTPServerRequest request, string _packname, User _user)
+	{
+		enforceUserPackage(_user, _packname);
+		const FilePart logo = request.files.get("logo");
+		enforceBadRequest(logo != FilePart.init);
+		auto renamed = NativePath.fromString(logo.tempPath.toString ~ logo.filename.name.extension);
```
wilzbach: Check file extensions here?

source/dubregistry/internal/utils.d
```diff
@@ -13,3 +26,42 @@ string black(string url)
...
+ *   file = the file to convert
+ *   deleteFinish = if true, delete the input file after a successful conversion
+ * Returns: the PNG stream of the icon or empty on failure
+ * Throws: Exception if name is empty or logo already exists and deleteExisting is not true
+ */
+bdata_t generateLogo(NativePath file) @safe
+{
+	static assert (isWeaklyIsolated!(typeof(&generateLogoUnsafe)));
+	static assert (isWeaklyIsolated!NativePath);
+	return (() @trusted => async(&generateLogoUnsafe, file).getResult())();
```
wilzbach: This should to timeout after X seconds and terminate the `convert` process (at the moment it doesn't)

source/dubregistry/dbcontroller.d
WebFreak001: this is `$unset`, not `$set`

wilzbach: Yes and I don't mind that you use `1`.
It's just more common to use `0` or `-1`

(Of course that mongodb itself promotes such a crap API is a different story)

WebFreak001: hm just took the first result off google when searching for how to unset

WebFreak001:
@wilzbach the issue with the hanging was actually an issue because of the code, fixed it now and svgs should work now, but because now we are restricting to png, bmp, jpg and gif it wont

views/my_packages.package.dt
```diff
@@ -19,6 +19,21 @@ block body
...
+	h2 General
+
+	.inputForm
+		form(method="POST", enctype="multipart/form-data", action="#{req.rootDir}my_packages/#{packageName}/set_logo")
+			h3 Logo
+			- string logoPath = "/packages/" ~ packageName ~ "/logo";
+			img.packageLogo(src=logoPath)
+			p.logoError.error
+			label
+				| Select new Logo (at most 1 MiB, optimal size 512x512)
```
wilzbach: Nit: typically `max.` is used for this (or not displayed at all except for the error case)

source/dubregistry/internal/utils.d
```diff
@@ -13,3 +28,78 @@ string black(string url)
...
+	(() @trusted {
+		foreach (chunk; png.stdout.byChunk(4096))
+			a.put(chunk);
+	})();
+
+	auto result = png.pid.wait;
+	timer.stop();
+	if (result != 0)
+	{
+		if (!timer.pending)
```
wilzbach: You manually stop the timer before, so this can never happen.
While this shouldn't be a problem anymore now that only a few image formats and sizes are allowed, ideally such timing code should be put in `utils`, s.t. it's easier to abstract / test / reuse it (or part of `vibe-core`).
For now the `timeout` from `coreutils` might be simpler and less error-prone to use?

source/dubregistry/internal/utils.d
```diff
@@ -13,3 +28,78 @@ string black(string url)
...
+	import std.array : appender;
+
+	// TODO: replace imagemagick command line tools with something like imageformats on dub
+
+	// use [0] to only get first frame in gifs, has no effect on static images.
+	string firstFrame = file.toNativeString ~ "[0]";
+
+	auto sizeInfo = execute(["identify", "-format", "%w %h %m", firstFrame]);
+	if (sizeInfo.status != 0)
+		return new LogoGenerateResponse(null, "Malformed image.");
```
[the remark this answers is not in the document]
WebFreak001: I would throw here, but it was broken when I tried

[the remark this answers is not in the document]
wilzbach: Yeah I'm aware - `Future` doesn't really support exceptions like es6 promises.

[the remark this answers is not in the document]
s-ludwig: And it also can't, at least in a safe way, because of their thread-local nature in D, especially since the latest RC additions. It should require the callback to be nothrow, though.

views/my_packages.package.dt
```diff
@@ -112,3 +127,23 @@ block body
...
 		form(method="POST", action="#{req.rootDir}my_packages/#{packageName}/remove")
 			button.danger(type="submit") Remove this package
+
+	:javascript
+		function updateLogoPreview(input) {
+			var f = input.files[0];
+			if (!f) return;
+			document.querySelector(".logoError").textContent = "";
+			if (f.type != "image/png" && f.type != "image/gif" && f.type != "image/jpeg" && f.type != "image/bmp")
+				document.querySelector(".logoError").textContent += "Warning: Invalid image type. ";
```
wilzbach: Oh I meant to use `return` to avoid creating the object url and displaying something that's invalid.

WebFreak001: I used this because it signals the user that the image is now broken instead of showing the valid previous image.

s-ludwig:
So, let's pull the trigger!

BTW, I made some package graphics back when designing the new DUB logo. The logo on the left side is not right, but this could be adjusted to serve as a more refined package logo later on:
dub-package

WebFreak001:
yeah that looks pretty cool, but I can imagine that it won't look good as small icon. I designed my icon so that it still looks good and recognizable even at 16x16. Though for large icons that could be cool. Maybe use the other D logo now that they are considering to rebrand to for the right side? Also I would remove or redo the shadows at the bottom

s-ludwig:
I don't think we should ever display the project logo that small, because who knows how the custom ones look like. But due to the coarse red strips it actually looks pretty decent in tiny sizes: dub-package-16 (the "D" could be made brighter at that size to stay readable)

True with the shadows. they need a frame or more space to the bottom to work like this. I missed the rebranding topic, do you have a reference?

WebFreak001:
uh I think I asked in #offtopic on slack once if the D logo there on https://github.com/dlang-community/artwork is the new D logo and they are planning on doing this, though it still has problems with small sizes like icons.

s-ludwig:
Ah okay, I see. My feelings about using it as the language logo are a little bit mixed, although it definitely is an awesome logo in general. Just as a side note, it could also make sense to go for a more gentle redesign, such as separating the moons and adjusting the shape of the "D". For example regarding the moons:

d-logo

views/view_package.dt
WebFreak001: @wilzbach oh actually I just found out that if a `packageName` were to be empty, this method would crash because split and splitter return an empty array for an empty string...

100 comments on this pr :tada:

wilzbach: Yeah, but that's not going to happen, is it?
Anyhow, if you are experiencing this problem somewhere else, there you might like `orElse`

https://github.com/dlang/phobos/pull/5154

I'm also a big fan or `safeDeref`, but I'm not sure whether the version on the NG supports ranges out of the box.

MartinNowak:
Nice to see this merged :).

I'd still say we should seed all package logos with the homepage/repo's `og:image` instead of adding a default logo. It simply reduces the necessary initial configuration for package maintainers.

PR #289, opened 2018-02-28
Remove mdondorff from authors

mdondorff:
I haven't been doing any noteworthy thing for dub-registry, hence removing myself from the 'Authors' section.


s-ludwig:
We should probably add at least the top 6 contributors there, somehow these author lists are even worse in that respect than copyright dates, they almost never get updated properly.

PR #291, opened 2018-02-28
immediate manual package update

MartinNowak:
- super annoying to wait half an hour for a new git tag
- no point in scheduling manual updates behind automated updates
- clicking trigger manual update usually means I just published sth. and need
  it, automated background checks have much lower priority
- chances of abuse are fairly low IMO and the cost to update a package is fairly
  low anyhow


views/my_packages.package.dt
```diff
@@ -31,18 +31,18 @@ block body
 		| #[strong= stats.total.to!string] downloads total
 
 	h2 Update
-	p Packages are automatically updated, so there is no need to trigger an update manually. Please, be patient.
```
wilzbach: :+1:

s-ludwig:
> no point in scheduling manual updates behind automated updates

The point is to guarantee a fair scheduling and to make it impossible to use this as a simple DoS vector. The chances for an attack may be low, but *if* they happen, it will be annoying having to implement counter-measures in real-time. The approach made a lot of sense before the polling intervals got so long. with a 5min polling duration, the likelihood that the queue is empty at a certain point in time is pretty high, and even if not, 5 minutes was barely a problem.

However, with the current situation, it makes sense to change the behavior as a quick workaround.

source/dubregistry/internal/workqueue.d
```diff
@@ -49,13 +50,35 @@ final class PackageWorkQueue {
 		}
 	}
 
+	void putFront(string pack_name)
```
s-ludwig: Why not implement a proper `putFront` in `FixedRingBuffer` instead?

MartinNowak: Because it takes ages until it's available here ;).

s-ludwig: Good point ;) I'll probably tag a hotfix release pretty shortly, though, to fix some issues that surfaced in conjunction with the vibe-core switch. There is just debugging of one source for network related slowdowns left.

wilzbach:
0.8.4-alpha.1 is finally available so I bumped the versions and kept only the first bit of the `static if`.

source/dubregistry/internal/workqueue.d
```diff
@@ -49,13 +50,27 @@ final class PackageWorkQueue {
...
+	{
+		synchronized (m_mutex) {
+			m_queue.putFront(pack_name);
```
wilzbach: In the first commit (https://github.com/dlang/dub-registry/pull/291/commits/28ba1327e40afe318c1baf53760dd60444af68b9), this used to do search for the package in the buffer and swap it to the front if it can be found. As Martin already used a `static if (is(typeof({ m_queue.putFront(pack_name); })))` blocked, I assumed that this was a preparation for the new Vibe.d version and just kept this block.
(This might lead to a subsequent update if a manual update was requested.)

s-ludwig: Maybe we should at least check for `m_queue[].count(pack_name) < 2`, so that spamming the update queue infinitely is not possible as easily.

WebFreak001: why would you want to update it multiple times? just the canFind as below would make sense here imo

s-ludwig: Just `canFind` would break the initial motivation for this, namely to be able to trigger an immediate update, regardless of the state of the general update queue. Ideally, if scheduled, it should be removed from its current spot and be put at the front instead, but having it be updated a second time during the normal update cycle still sounds completely reasonable.

WebFreak001: oh yeah, hm I think moving it would make more sense than updating it twice, all these git providers have a rate limit after all

s-ludwig: Although the rate-limit for GitHub only applies for non-cached responses, so we are generally fine there. We could also stretch out the regular update interval to an hour to reduce the general stress a bit.

wilzbach: > Maybe we should at least check for m_queue[].count(pack_name) < 2, so that spamming the update queue infinitely is not possible as easily.

Went with `if (!m_queue[0 .. min(10, $)].canFind(pack_name))` - seems a bit more robust, but it's still "equally naive".
A proper protection would probably check that a package isn't updated/triggered more than X times within one hour ...

PR #293, opened 2018-03-05
Added cms under web-related.

bausshf:


categories.json
```diff
@@ -63,6 +63,7 @@
 				{"name": "auth", "description": "Authentication and user management"},
 				{"name": "communication", "description": "Communication components"},
 				{"name": "framework", "description": "Web framework"}
```
s-ludwig: Missing a comma here.

s-ludwig:
Just to make sure, is content management system *library* correct, or is it an application?

bausshf:
Well I guess it would be fine as library.

PR #301, opened 2018-03-25
Make a few performance tweaks to get closer to a more reasonale memory usage

s-ludwig:


s-ludwig:
@MartinNowak Do you think that there is still value in `recomputeScores` at all? I guess it should be sufficient to let the first round of statistics updates recompute the score instead.

source/dubregistry/dbcontroller.d
```diff
@@ -183,27 +195,32 @@ class DbController {
 
 		if (png.length)
 		{
-			m_packages.update(["name": packname], ["$set": [
-				"logo": BsonBinData(BsonBinData.Type.generic, png),
-				"logoHash": BsonBinData(BsonBinData.Type.md5, cast(bdata_t)md5Of(png)[].idup)
-			]]);
+			BsonBinData hash = BsonBinData(BsonBinData.Type.md5, cast(bdata_t)md5Of(png)[].idup);
```
WebFreak001: actually now that all extra files are going to be stored in that collection this shouldn't be md5 anymore, the hash here was only used for updating cache on browsers after a logo change. Something longer like sha512 should be used here instead considering we already have over 1000 packages and I am planning on actually storing some more files per package in my next PR. While there won't be any problem right now and probably not in the next few years or even decades, we should rather go a bit safer for the future.

s-ludwig: A perfect 128-bit hash has a collision probability somewhere around 10^-27 for 100 billion different files, which is still virtually non-existant. A 160-bit SHA-1 would be better, but 512-bit seems to be way over the top for this purpose (remember that git works with SHA1 and UUIDS are also just 128-bit).

The reason why I went with the existing MD5 value was mainly because BSON has a dedicated value type for this, which may allow the `BsonBinData` to be treated as a fixed-size entity later on.

With MD5 there is in theory the risk of a pre-image attack to infiltrate foreign packages with a malicious logo graphic (doesn't seem possible in practice due to the ImageMagick conversion in-between, though). But to be a bit safer we could use a UUID v5 (truncated SHA-1) and still store it as a special `BsonBinData`.

Otherwise, another simple possibility would of couse be to use a plain `BsonObjectID` that is newly generated for each uploaded logo.

WebFreak001: yeah just using the builtin object id would be fitting and still fulfill the task the logoHash had before and seems impossible to collide for this purpose.

source/dubregistry/registry.d
```diff
@@ -225,7 +225,7 @@ class DubRegistry {
 		nfo["dateAdded"] = pack._id.timeStamp.toISOExtString();
 		nfo["owner"] = pack.owner.toString();
 		nfo["name"] = pack.name;
-		nfo["logoHash"] = pack.logoHash.rawData.toHexString;
+		nfo["logo"] = pack.logo.toString;
```
WebFreak001: this is getting really confusing now, can you rename the new field to logoID instead?

s-ludwig:
Hm, didn't mean to rebase the whole PR - only the last two commits have changed. So I've removed the "logo" field completely from the `Json` information and instead added a statically typed field to `PackageInfo` now. This has the advantage that we don't needlessly leak the logo ID to the public API.

During testing I also noticed that errors and exceptions during the logo upload process are not yet behaving as expected and added another commit to fix that, too.

Issue #160, opened 2016-05-27
show stats as graph


wilzbach:
I really like the stats at the sublime registry:

![image](https://cloud.githubusercontent.com/assets/4370550/15611470/c6d3550c-242a-11e6-918d-2d7798a847cf.png)
https://packagecontrol.io/packages/SublimeLinter-contrib-stylelint

As we already have the API, we could just show a similar graph with a bit of JS?

https://code.dlang.org/api/packages/mir/stats

wilzbach:
I just saw that Rust provides a similar nice overview:

![image](https://cloud.githubusercontent.com/assets/4370550/15628302/f205977a-24fd-11e6-9da6-f93ce8bb5e2e.png)

https://crates.io/crates/abomonation

MartinNowak:
Do we want server-side or client-side images? We don't even have jQuery atm.
Would be a fun opportunity to check out [ggplotd](http://code.dlang.org/packages/ggplotd).

wilzbach:
> Do we want server-side or client-side images? >We don't even have jQuery atm.
>Would be a fun opportunity to check out ggplotd.

You probably want to support all devices nicely, so there isn't a good way around JavaScript.
Also people love interactivity...

If you worry about the performance impact, the plotting library will be cached on the second time a user requests it, but I actually don't think that something like Plotly or c3 will be a big deal... Especially if it's not put in the HTML header and thus blocking the first page load.

ZombineDev:
+1 for using client-side charts (as much as I dislike JS). I used to work for UI controls company, and I can tell you from experience that doing any server-side charting besides coarse-grained data processing has only downsides:
* Very bad performance:
  * Server-side rendering is very CPU and network intensive (assuming the result is a an image). The results can be cached, but this means almost no interactivity.
    * most client-side UI events require a request to the server. In addition to bad UX, this increases the load on the server.
    * Some combinations of chart features can expose performance bottlenecks, which opens  serface for DoS attack.
* Bad UX
    * Raster graphics are not easily adaptable to different resolutions / form factors
    * Anti-aliasing is not practical for performance reasons.
* etc.

MartinNowak:
Would be my gut feeling as well @ZombineDev.
Any suggestions other than highcharts.com?

s-ludwig:
Well, if anything server side, I'd output SVG, which does not have those issues. But I don't mind client-side either.

wilzbach:
> Any suggestions other than https://highcharts.com?

[...]

PR #302, opened 2018-03-25
package subtabs + doc link + more spacing + better readme

WebFreak001:
readme is now detected from repo content instead of hardcoding README.md
bitbucket api only tested using external debugging of http calls
gitlab not tested at all due to missing api keys
github works fine
external documentation link to dpldocs.info (asked @adamdruppe for this)
planning support for sample file detection here (using repo content)

part of #265, rewritten and optimized a bit

showing dub registry page in large

---

showing dub registry page in small

---

showing dub registry page on mobile


views/view_package.sample.dt
```diff
@@ -0,0 +1 @@
+p TODO: render sample #{sample}
```
wilzbach: Maybe move to a follow-up PR?

views/view_package.dt
```diff
@@ -124,10 +150,21 @@ block body
 			- if( auto pl = "copyright" in versionInfo )
 				li#copyright= pl.get!string
 
+		- if (isMine)
+			//- TODO: make .actions not conditional on isMine, instead have actions which are always there here too (either as a: button.large for actions or p.detail for descriptions)
```
wilzbach: Does this actually make sense to display a greyed "Manage this package" button?

views/view_package.dt
```diff
@@ -87,20 +93,40 @@ block body
...
+					- msettings.urlFilter = &urlFilter;
+					!= filterMarkdown(readmeContents, msettings)
+			- else
+				.repositoryReadme
+			- renderedTabContent = true;
+		- else
+			- foreach (i, sample; sampleURLs)
+				- if (activeTab == "sample_"~i.to!string)
+					- renderedTabContent = true;
+					include view_package.sample
```
wilzbach: Maybe move this in a separate PR, s.t. documentation button can be rolled out before this feature is complete?

views/view_package.dt
```diff
@@ -87,20 +93,40 @@ block body
...
+				.repositoryReadme.markdown-body
+					- import vibe.textfilter.markdown : MarkdownFlags, MarkdownSettings, filterMarkdown;
+					- scope msettings = new MarkdownSettings;
+					- msettings.flags = MarkdownFlags.backtickCodeBlocks|MarkdownFlags.noInlineHtml|MarkdownFlags.tables;
+					- msettings.headingBaseLevel = 2;
+					- msettings.urlFilter = &urlFilter;
+					!= filterMarkdown(readmeContents, msettings)
+			- else
+				.repositoryReadme
+			- renderedTabContent = true;
```
wilzbach: Another tab we could add would be stats e.g. like this:
https://github.com/dlang/dub-registry/issues/160

views/view_package.dt
WebFreak001: no not a grayed "Manage this package" button, the button should only be visible if logged in and as the author, but the `.actions` div there should be always visible once we have another action

views/view_package.dt
WebFreak001: if we go through with the review process quickly we can roll this out early too ;)

views/view_package.sample.dt
WebFreak001: was planned

views/view_package.dt
wilzbach: Another one could be versions: http://code.dlang.org/packages/vibe-d/versions

adamdruppe:
i like it

WebFreak001:
arsd actually has a readme now after this PR

FraMecca:
I like it a lot

p0nce:
Obviously better

source/dubregistry/registry.d
```diff
@@ -408,8 +409,31 @@ class DubRegistry {
...
+				// check for all other readmes such as README.txt, README.jp.md, etc.
+				readme = files.countUntil!(a => a.type == RepositoryFile.Type.file && a.path.head.name.asUpperCase.startsWith("README"));
+			}
+
+			if (readme != -1) {
+				rep.readFile(reference.sha, files[readme].path, (scope input) {
+					dbver.readme = input.readAllUTF8();
+					string ext = files[readme].path.head.name;
+					// endsWith doesn't like to work with asLowerCase
+					dbver.readmeMarkdown = ext.endsWith(".md") || ext.endsWith(".MD") || ext.endsWith(".Md") || ext.endsWith(".mD");
```
wilzbach: Yeah the problems is that `asLowerCase` doesn't yield a bidirectional range :/
A workaround is this:

```
"foo.mD".retro.asLowerCase.startsWith(".md".retro)
````

https://run.dlang.io/is/PAuvw4

Though it's not very nice either.
Another one:

```
alias lowerCasePred = (a, b) => a.only.sicmp(b.only) == 0;
writeln("foo.mD".endsWith!lowerCasePred(".md"));
```

https://run.dlang.io/is/gdhzFQ

Comment (WebFreak001) [reply source/dubregistry/registry.d]: uh any problem with this? It's just 4 checks

Comment (s-ludwig) [reply source/dubregistry/registry.d]: Or just `ext.extension.icmp(".md") == 0`.

Comment (WebFreak001) [reply source/dubregistry/registry.d]: well probably not icmp though, we don't want combined unicode characters in the filenames

**Comment by s-ludwig** [comment]
Sounds good.

**Comment by s-ludwig** [comment]
Now also deployed.

**Comment by MartinNowak** [comment]
The doc link is visually confusion, it looks like a tab but it's an external link.

I've also argued a bit against centralized and automatically build documentation for two reasons, reliability of the service [¹] and reservations against that being a proper solution for sloppy or lacking documentation [²].

[¹]: https://forum.dlang.org/post/p7d0c2$8h5$1@digitalmars.com
[²]: https://forum.dlang.org/post/p7e79k$13cp$1@digitalmars.com

It might at least encourage people to fix their docs, so let's see how that works out.
We might also want to have overridable doc links, so that people don't have to maintain 2 doc systems.

**Comment by MartinNowak** [comment]
Moving forward with extracting metadata from unstructured repos and readmes seems like a maitenance intensive way forward. I'd suggest to rather support enabling techniques, such as relative links in READMEs, that would also work across github and code.dlang.org, similarly to the documentation link.

File: public/styles/common.css [hunk 6]
```diff
@@ -172,12 +172,88 @@ h3 {
...
+	border-top-left-radius: 4px;
+	border-top-right-radius: 4px;
+	margin-right: 4pt;
+}
+
+.subtabsHeader > .tab.active, .subtabsHeader > .tab:hover, .subtabsHeader > .tab:focus {
+	background-color: #98312A;
+}
+
+.subtabsHeader > .tab.external::after {
```
MartinNowak: tab external is quite a contradiction ;)

source/dubregistry/registry.d
```diff
@@ -408,8 +411,31 @@ class DubRegistry {
...
+			// check exactly for readme.me
+			ptrdiff_t readme;
+			readme = files.countUntil!(a => a.type == RepositoryFile.Type.file && a.path.head.name.asUpperCase.equal("README.MD"));
+			if (readme == -1) {
+				// check exactly for readme
+				readme = files.countUntil!(a => a.type == RepositoryFile.Type.file && a.path.head.name.asUpperCase.equal("README"));
+			}
+			if (readme == -1) {
+				// check for all other readmes such as README.txt, README.jp.md, etc.
+				readme = files.countUntil!(a => a.type == RepositoryFile.Type.file && a.path.head.name.asUpperCase.startsWith("README"));
```
MartinNowak: Nice, somewhat problematic that we cannot render them though.

WebFreak001:
relative links to other MD files sounds like a great idea.

Though I don't see how extracting documentation links from the readme is going to be a maintenance effort. Support the basic and most common ways that people link their docs with so we have like 80-90% of repos covered, and the others can set it via a new input text field which will eventually get added.

source/dubregistry/registry.d
WebFreak001: well at least support for README, Readme, readme, ReadMe, etc now. i18n is something we might eventually look into, I have seen multiple japanese repositories with different readme names with this README.en.md pattern

MartinNowak:
> relative links to other MD files sounds like a great idea.

Relative links to anything, including source samples, just like on GH.

> Though I don't see how extracting documentation links from the readme is going to be a maintenance effort.

Cause the info needs to be maintained for GH and be extractable for code.dlang.org metadata parser. Basically README.md already allows people to do anything they want with their page, though indeed some guidance on a good format would be nice.
Typically samples are included in README and full file examples could be linked from their, don't see why we should invent our own sample_ convention and lift the info to the code.dlang.org layout.

MartinNowak:
Very nice changes otherwise btw :).

MartinNowak:
Also client-side tabbing and replacing query params `?tab=info` with anchors `#readme` would remain compatible with other html layouts, e.g. scrolling.

PR #303, opened 2018-03-26
Add Top5 {most popular,created,update} front page

wilzbach:
As we don't seem to make progress on that one, I hacked a "top10" front page together:

Desktop:

image

Mobile:

image

It's far from perfect, but I hope this gets the discussion started.

### DIFF
--- a/source/dubregistry/web.d
+++ b/source/dubregistry/web.d
@@ -70,22 +70,10 @@ class DubRegistryWebFrontend {
 		setTimer(30.seconds, &updatePackageList, true);
 	}
 
-	@path("/")
-	void getHome(string sort = "updated", string category = null, ulong skip = 0, ulong limit = 20)
+	private auto searchForPackages(string sort = "updated", string category = null, ulong skip = 0, ulong limit = 20)
 	{
 		import std.algorithm.comparison : min;
 		import std.algorithm.iteration : filter, map;
-
-		static struct Info {
-			// TODO: No need to use untyped Json
-			static struct Package { DbPackageStats stats; Json _; alias _ this; }
-			Package[] packages;
-			size_t packageCount;
-			ulong skip;
-			ulong limit;
-			Category[] categories;
-			Category[string] categoryMap;
-		}
 
 		static import std.algorithm.sorting;
 		import std.algorithm.searching : any;
@@ -111,14 +99,16 @@ class DubRegistryWebFrontend {
 			return getDate(a) > getDate(b);
 		}
 		switch (sort) {
-			default: std.algorithm.sorting.sort!compare(packages); break;
 			case "name": std.algorithm.sorting.sort!((a, b) => a.name < b.name)(packages); break;
 			case "score": std.algorithm.sorting.sort!((a, b) => a.stats.score > b.stats.score)(packages); break;
 			case "added": std.algorithm.sorting.sort!((a, b) => a.dateAdded > b.dateAdded)(packages); break;
+			case "updated":
+			default: std.algorithm.sorting.sort!compare(packages); break;
 		}
+		// TODO: No need to use untyped Json
+		static struct Package { DbPackageStats stats; Json _; alias _ this; }
 
 		// limit package list to current page
-		size_t pcnt = packages.length;
 		packages = packages[min(skip, $) .. min(skip + limit, $)];
 
 		// collect package infos
@@ -126,17 +116,58 @@ class DubRegistryWebFrontend {
 		foreach (p; m_registry.getPackageInfos(packages.map!(p => p.name).array))
 			infos[p.info["name"].opt!string] = p.info;
 
-		Info info;
-		info.packageCount = pcnt;
-		info.packages = packages
-			.map!(p => Info.Package(p.stats, infos.get(p.name, Json.init)))
-			.array;
-		info.skip = skip;
-		info.limit = limit;
-		info.categories = m_categories;
-		info.categoryMap = m_categoryMap;
+		return packages.map!(p => Package(p.stats, infos.get(p.name, Json.init))).array;
+
+	}
+
+	void search(string sort = "updated", string category = null, ulong skip = 0, ulong limit = 20)
+	{
+		auto packages = searchForPackages(sort, category, skip, limit);
+
+		static struct Info {
+			typeof(searchForPackages()) packages;
+			size_t packageCount;
+			ulong skip;
+			ulong limit;
+			Category[] categories;
+			Category[string] categoryMap;
+		}
+
+		Info info = {
+			packageCount: packages.length,
+			packages: packages,
+			skip: skip,
+			limit: limit,
+			categories: m_categories,
+			categoryMap: m_categoryMap,
+		};
 
 		render!("home.dt", info);
+	}
+
+
+	@path("/")
+	void getHome(string sort = "updated", string category = null, ulong skip = 0, ulong limit = 20)
+	{
+		if (request.requestURI != "/")
+			return search(sort, category, skip, limit);
+
+		limit = 5;
+		auto topScored = searchForPackages("score", null, 0, limit);
+		auto topAdded = searchForPackages("added", null, 0, limit);
+		auto topUpdated = searchForPackages("updated", null, 0, limit);
+
+		static struct Info {
+			size_t packageCount;
+			Category[] categories;
+			Category[string] categoryMap;
+		}
+		Info info = {
+			packageCount: m_packages.length,
+			categories: m_categories,
+			categoryMap: m_categoryMap,
+		};
+		render!("start.dt", topScored, topAdded, topUpdated, info);
 	}
 
 	// compatibility route
--- a/views/start.dt
+++ b/views/start.dt
@@ -1,0 +1,82 @@
+extends layout
+
+block title
+	- import dubregistry.viewutils;
+	- import dubregistry.web;
+	- import vibe.data.json;
+	- import std.algorithm.iteration : map;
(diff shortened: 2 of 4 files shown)


views/start.dt
```diff
@@ -0,0 +1,98 @@
...
+							- foreach (c; cat ? cat.subCategories : info.categories)
+								- outputCat(c);
+					- outputCat(null);
+				button#category-submit(type="submit") Update
+		#category-dynamic-form(style="display: none")
+			p Select category:
+				- foreach (i; 0 .. 6)
+					select(id="categories_#{i}", name="categories_#{i}", onChange='setCategoryFromSelector(#{i})')
+			:javascript
+				setupCategoryForm();
```
wilzbach: This is just copied from `home.dt`, probably on the long run this should be replaced with keywords.

views/start.dt
```diff
@@ -0,0 +1,98 @@
...
+										- imageDescription = pc.imageDescription;
+								- if (!image.length) image = "unknown";
+								- if (!imageDescription.length) imageDescription = "unknown";
+								span(class="category-icon icon-category-#{image} #{isApplication ? `application` : ``}", title=imageDescription)
+							.front-package-name
+								a(href='packages/#{pl["name"].get!string}')= pl["name"].get!string
+							.front-package-version
+								| (#{ ver.length <= 20 ? ver : ver[0 .. 18] ~ "&hellip;" })
+							//-.front-package-date
+								//-|= formatFuzzyDate(p["date"])
```
wilzbach: Should we display the date?

views/start.dt
```diff
@@ -0,0 +1,98 @@
...
+									- foreach (category; pl["categories"])
+										- if (category.get!string.startsWith("application"))
+											- isApplication = true;
+											- break;
+									- if (auto pc = pl["categories"][0].get!string in info.categoryMap)
+										- image = pc.imageName;
+										- imageDescription = pc.imageDescription;
+								- if (!image.length) image = "unknown";
+								- if (!imageDescription.length) imageDescription = "unknown";
+								span(class="category-icon icon-category-#{image} #{isApplication ? `application` : ``}", title=imageDescription)
```
wilzbach: Should the logo be bigger? (maybe a good idea - especially with the new logo upload functionality)

wilzbach: Also we should probably refactor this to a `getLogo` method for the Packages

source/dubregistry/web.d
```diff
@@ -97,25 +85,69 @@ class DubRegistryWebFrontend {
...
-		Info info;
-		info.packageCount = packages.length;
-		info.packages = packages[min(skip, $) .. min(skip + limit, $)]
-			.map!(p => Info.Package(p.stats, m_registry.getPackageInfo(p, false).info))
-			.array;
-		info.skip = skip;
-		info.limit = limit;
-		info.categories = m_categories;
-		info.categoryMap = m_categoryMap;
+	void search(string sort = "updated", string category = null, ulong skip = 0, ulong limit = 20)
```
wilzbach: Can't be private due to the use of `render` :/

wilzbach:
Another cool metric would be "Most depended upon", but that's a bit more difficult, so I would reserve this for another PR.

WebFreak001:
why don't you add the package logo to that? Those card renders of a package were a perfect place for logos when I thought of the feature. Maybe make a generic diet function for rendering such a card too? I am planning on adding a per user package view where this would be useful.

s-ludwig:
My first impression was that there is still far too much going on on the page. Maybe this will be solved by tweaking the style to better structure the lists (especially to separate individual items), or maybe the number of entries needs to be reduced to more like 8.

I had to double-check the red headings and would favor a pure text heading with a separate "view all..." link at the bottom, since headings that also function as links are rather unusual and a bit ambiguous if not augmented with additional visual cues. That part of the page also appears a bit too reddish, the red should rather be reserved as more of an accent color.

s-ludwig:
This is the visual direction that I'd go, based on what we already got over the last months:
grafik

Not sure if we need to make the exact style part of the PR, though.

WebFreak001:
+1, I think a basic style should be included though or nobody will ever do it :p

Also don't stay too close to that draft, the font and shadows are not consistent with the site as of now ;)

But at least consistent height for the cards would be nice, maybe shadows are a bit too much here because of how many there are, just some borders between them would be enough imo for the current style.

s-ludwig:
Yeah, I didn't touch the pre-selected font when mocking up that graphic, although the font on the site is something that I really dislike but never had the energy to attempt to change.

W.r.t. shadows and lines, I'm aiming for a more open style that allows things to breathe a bit and would rather reduce the number of items than to go more in a list/table direction. After all this is just meant to give a friendly overview of the current state, not to show an exhaustive list in a form that aims for space efficiency.

wilzbach:
@s-ludwig that looks really nice!

Did you already do this in CSS or was it a pure mockup. If so feel free to push to this PR, so that Jan and I can improve upon it/play with it?

s-ludwig:
It was just a pure mockup, but the card itself would have a style close to this one:

	.card {
		border-radius: 12px;
		box-shadow: 0 5px 18px #d7d3d1;
		padding: 14px;
	}

The icon is 64x64 px, with another 14px margin to the right

WebFreak001:
maybe also add a `border-radius: 6px;` to the img tag then, I would actually use static positioning for the layouting here with 3 core parts: img, title+tag, description (`line-height: 1.2em; height: 2.4em;`)

wilzbach:
> It was just a pure mockup, but the card itself would have a style close to this one:

Nice. Thanks!
I played a bit further:

image

image

> maybe also add a border-radius: 6px; to the img tag then

I only used the default logo for simplicity for now.

views/start.dt
```diff
@@ -0,0 +1,85 @@
...
+			ul
+				- foreach (pl; packages)
+					- if( pl["versions"].length )
+						- auto vidx = getBestVersionIndex(pl["versions"].get!(Json[]).map!(v => v["version"].opt!string));
+						- auto p = pl["versions"][vidx];
+						- auto desc = p["description"].opt!string;
+						- auto ver = p["version"].opt!string();
+						li
+							.front-package-logo
+								img(src="images/default-logo.png")
```
WebFreak001: why did you hardcode the default logo here instead of using the per logo endpoint? `/packages/<package>/logo`

views/start.dt
```diff
@@ -0,0 +1,85 @@
...
+					- outputCat(null);
+				button#category-submit(type="submit") Update
+		#category-dynamic-form(style="display: none")
+			p Select category:
+				- foreach (i; 0 .. 6)
+					select(id="categories_#{i}", name="categories_#{i}", onChange='setCategoryFromSelector(#{i})')
+			:javascript
+				setupCategoryForm();
+
+		- void listPackages(P)(P packages) {
```
WebFreak001: is the `{` neccessary here? I know that it works with normal functions (not template functions)

views/start.dt
```diff
@@ -0,0 +1,85 @@
...
+			ul
```
WebFreak001: make this ul contain a classname and make the css apply to that `ul.packageList` or similar instead of hardcoding it just for the homepage so it can be reused in other places.

WebFreak001:
hm can you add a little bit more horizontal padding (like 2-4pt on both sides) inside the version tags? They look a bit squished

WebFreak001:
otherwise looks good to me

jacob-carlborg:
Why is `vibe-d` listed twice?

wilzbach:
> Why is vibe-d listed twice?

Well apparently vibe-d is popular and frequently updated. We could filter for this (on the current preview `libdparse` is popular + updated), but for now I have the opinion that before we complicate the code, we should figure out whether this is really unwanted.
After all, this is just an initial take at improving the front page, we can always do better.
At the moment I'm really just interested in getting an initial shot at a more modern landing page.

BTW @WebFreak001 what do you think about re-using the cards/layout defined in this PR for the search results too?
Might be a way too "show-off" the new logos.

s-ludwig:
W.r.t. to the information in the corners of the cards, I think this regressed a bit too far in terms of information load. I'd try to keep the version tag aligned with the package name (since that is already known from the package page and it avoids the need to scan the whole width of the card), and reduce the status text at the bottom to only a single side (maybe just drop the registrant name for at least the second and third columns).

The "x days ago" should then also be column specific (e.g. "added 3 days ago" and "update 2 hours ago"). Not sure about the first column, maybe just the author name there?

s-ludwig:
The slightly unfortunate effect of the status line is also that the cards are not symmetric around the logo anymore, which adds a lot to the perceived visual simplicity.

WebFreak001:
re-using cards for search results sounds like a great idea, in my PR after #302 I will also add user profile package lists with these cards and extend the home page you have basically created here like in #265

WebFreak001:
last one looks best and least confusing imo

bausshf:
@s-ludwig People are usually more interested in the latest stable version anyway

s-ludwig:
Exactly. For the "recently updated" column, it could still be interesting to see pre-release updates, but as a default, the stable version looks like the best choice.

bausshf:
Perhaps the preference could be a user-setting

WebFreak001:
---

:ballot_box_with_check: Show pre-release versions

---

MartinNowak:
Really cool stuff, glad to see this finally in.

Next :wink:.
- refactoring the sorted home list to be identical to an empty search
- replacing categories with tags (just categories.spit('.'))
- a typeahead search box (with search words and tags)

PR #306, opened 2018-03-26
Only store data necessary for sorting the package list in memory.

s-ludwig:
This removes a source of high memory use (~80MB) and also a source for what appears as a memory leak. It's not clear what keeps the memory pinned for the old approach, but it showed a steady increase in memory consumption for each package list update.

### DIFF
--- a/source/dubregistry/registry.d
+++ b/source/dubregistry/registry.d
@@ -215,12 +215,18 @@ class DubRegistry {
 
 	/** Gets information about multiple packages at once.
 
-		The order of the returned packages is undefined and some of the package
-		names may not result in a `PackageInfo` to be returned at all, if no
-		package with the given name exists, so that the returned range needs
-		to be manually matched against the list of names.
+		The order and count of packages returned may not correspond to the list
+		of supplied package names. Only those packages that actually reference
+		an existing package will yield a result element.
+
+		The consequence is that the caller must manually match the result to
+		the supplied package names.
 
 		This function requires only a single query to the database.
+
+		Returns:
+			An unordered input range of `PackageInfo` values is returned,
+			corresponding to all or part of the packages of the given names.
 	*/
 	auto getPackageInfos(scope string[] pack_names, bool include_errors = false)
 	{
--- a/source/dubregistry/registry.d
+++ b/source/dubregistry/registry.d
@@ -213,6 +213,21 @@ class DubRegistry {
 		return getPackageInfo(pack, include_errors);
 	}
 
+	/** Gets information about multiple packages at once.
+
+		The order of the returned packages is undefined and some of the package
+		names may not result in a `PackageInfo` to be returned at all, if no
+		package with the given name exists, so that the returned range needs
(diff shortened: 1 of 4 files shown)


source/dubregistry/registry.d
```diff
@@ -213,6 +213,21 @@ class DubRegistry {
 		return getPackageInfo(pack, include_errors);
 	}
 
+	/** Gets information about multiple packages at once.
+
+		The order of the returned packages is undefined and some of the package
+		names may not result in a `PackageInfo` to be returned at all, if no
+		package with the given name exists, so that the returned range needs
```
wilzbach: s/so that/so

source/dubregistry/web.d
```diff
@@ -100,13 +114,22 @@ class DubRegistryWebFrontend {
...
+		// collect package infos
+		Json[string] infos;
+		foreach (p; m_registry.getPackageInfos(packages.map!(p => p.name).array))
+			infos[p.info["name"].opt!string] = p.info;
+
 		Info info;
-		info.packageCount = packages.length;
-		info.packages = packages[min(skip, $) .. min(skip + limit, $)]
-			.map!(p => Info.Package(p.stats, m_registry.getPackageInfo(p, false).info))
+		info.packageCount = package_count;
```
wilzbach: Ehm mixing styles couldn't be more obvious ;-)

source/dubregistry/web.d
```diff
@@ -363,8 +386,28 @@ class DubRegistryWebFrontend {
 
 	private void updatePackageList()
 	{
-		setTimer(30.seconds, &updatePackageList);
```
wilzbach: I assume extending the update period from 30 seconds to the 30 minute watcher in app.d was done on purpose.

source/dubregistry/web.d
s-ludwig: You are right in so far that this should be `pcnt` or similar according to the vibe.d style guide. `package_count` would be the correct style for parameters, though.

source/dubregistry/web.d
s-ludwig: Do you mean that something was changed in this PR in that regard? It was 30 seconds and it still is (line 70). I just changed this to a periodic timer stead of starting a new one every time.

source/dubregistry/registry.d
s-ludwig: I think the "so that" is okay here, but the sentence is borderline in length anyway, so I'll reword that.

s-ludwig:
To be fair, I wouldn't really call this a "workaround" in the sense that the previous quick'n'dirty approach was really wasteful in terms of memory. But of course this whole piece of code is a workaround for a missing sorting implementation employed by the database directly. At some point we should add an additional sorting key for the latest version and get rid of this in-memory list.

source/dubregistry/web.d
wilzbach: Ah didn't see that. I was referring to the global timer.

PR #307, opened 2018-03-27
Use netlify for previewing PRs

wilzbach:
Looks like this approach is going to work :)

Their docker container is a bit limited, but no problem - we can download all binaries we need.

For now this just makes the a few pages visible.
Next steps:
- use the mirror mode to fetch all packages.
- find a workaround against filenames containing '?' (netlify doesn't support these)

BTW we could netlify too also deploy all the documentation pages to dub.pm directly from this repo (though we still might put them in a separate repo or sub-repo, s.t. the builds are faster here)

PR #308, opened 2018-03-29
 Improve the netlify preview to show packages

wilzbach:
As mentioned in https://github.com/dlang/dub-registry/pull/307, downloading the Json doesn't work (or well it takes 10 minutes), so I made `--mirror` support importing a manually downloaded copy of the Json file.

This is also handy for my local development, when I want to quickly nuke the entire database and on my machine `--mirror=...` doesn't work well either (could be due to my crappy internet though).

### DIFF
--- a/netlify.sh
+++ b/netlify.sh
@@ -7,6 +7,7 @@ set -x
 DMD_VERSION="2.079.0"
 BUILD_DIR="build"
 MONGO_VERSION="mongodb-linux-x86_64-ubuntu1404-3.6.3"
+CURL_FLAGS=(-fsSL --retry 10 --retry-delay 30 --retry-max-time 600 --connect-timeout 5 --speed-time 30 --speed-limit 1024)
 
 # install mongo
 if [ ! -f mongo.tgz ] ; then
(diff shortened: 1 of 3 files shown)


netlify.sh
```diff
@@ -21,29 +22,39 @@ PID_MONGO=$!
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd "${DIR}"
 
-. "$(curl -fsSL --retry 5 --retry-max-time 120 --connect-timeout 5 --speed-time 30 --speed-limit 1024 https://dlang.org/install.sh | bash -s install "dmd-${DMD_VERSION}" --activate)"
+. "$(curl "${CURL_FLAGS[@]}" https://dlang.org/install.sh | bash -s install "dmd-${DMD_VERSION}" --activate)"
+
+curl "${CURL_FLAGS[@]}" https://code.dlang.org/api/packages/dump | gunzip > mirror.json
```
wilzbach: We could also upload a JSON dump somewhere else if this becomes a problem, but for now there aren't too many PRs at dub-registry so having a separate file for this will probably just be more work.

PR #310, opened 2018-03-30
Add --no-monitoring for local development

wilzbach:
I get tons of unrelated errors when I run the full registry for development locally.
Also it's nice for local development to update your "dev dump" deliberately and not every X minutes.

Moreover, my IP anonoymous quota of GitHub API queries get exceeded super fast, s.t. other services that depend on it don't work anymore.

There also seems to be an issue with the Vibe.d, because it sometimes gets unresponsive/blocks when updates failed.

### DIFF
--- a/source/app.d
+++ b/source/app.d
@@ -44,12 +44,14 @@ version (linux) private immutable string certPath;
 
 shared static this()
 {
+	bool noMonitoring;
 	setLogFile("log.txt", LogLevel.diagnostic);
 
 	string hostname = "code.dlang.org";
 
 	readOption("mirror", &s_mirror, "URL of a package registry that this instance should mirror (WARNING: will overwrite local database!)");
 	readOption("hostname", &hostname, "Domain name of this instance (default: code.dlang.org)");
+	readOption("no-monitoring", &noMonitoring, "Don't periodically monitor for updates (for local development)");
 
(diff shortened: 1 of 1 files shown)


source/app.d
```diff
@@ -44,12 +44,14 @@ version (linux) private immutable string certPath;
 
 shared static this()
 {
+    bool noMonitoring;
```
WebFreak001: mixed tabs and spaces

source/app.d
```diff
@@ -80,7 +82,8 @@ shared static this()
 
 	auto router = new URLRouter;
 	if (s_mirror.length) router.any("*", (req, res) { req.params["mirror"] = s_mirror; });
-	router.get("*", (req, res) @trusted { if (!s_checkTask.running) startMonitoring(); });
+	if (!noMonitoring)
+	    router.get("*", (req, res) @trusted { if (!s_checkTask.running) startMonitoring(); });
```
WebFreak001: ditto

source/app.d
```diff
@@ -115,5 +118,6 @@ shared static this()
 	listenHTTP(settings, router);
 
 	// poll github for new project versions
-	startMonitoring();
+	if (!noMonitoring)
+	    startMonitoring();
```
WebFreak001: ditto

Issue #309, opened 2018-03-30
Have a CDN with all API responses statically generated


wilzbach:
Similar to the new Netlify preview, we could have a DUB mirror that updates itself daily (Travis cron) and is at the end of the chain and the last resort.

The idea: prefetch all responses with wget/curl and deploy them to a CDN (GitHub Pages, Netlify, ...).
Netlify can handle a few GBs quite fine (the docarchives.dlang.io currently take 3G), but I doubt that it 
will be that much. Currently the uncompressed JSON of all packages is 38M
Netlify does support redirects, so the zips don't need to be downloaded.

Downsides/problems:
- search won't work (ok, because this is last resort, s.t. building always works)

bausshf:
You could possibly make search work by generating a static javascript object, then search through that with javascript, instead of using an api call.

Of course that would be additional functionality to add on top.

MartinNowak:
I'd rather argue for migrating the registry to properly hosted servers and a simple HA setup instead of complicating the architecture with such a strong requirement (or uglifying it with a hack).
We could possibly funnel through cloudfront instead of using curl/wget, but I have bad experience with long 503s with unreliable backend servers.

In fact I'm preparing migration to a couple of VPS instances atm., let's try the proved approach first.

PR #311, opened 2018-03-30
[WIP] Experiment with generating all API data statically

wilzbach:
See https://github.com/dlang/dub-registry/issues/309

A start, probably `curl` is too slow and this should would need to be done directly in the registry.

TODO:
- [ ] generate a list of redirects (see https://www.netlify.com/docs/redirects/)


source/app.d
```diff
@@ -33,7 +33,7 @@ void startMonitoring()
 		sleep(1.seconds()); // give the cache a chance to warm up first
 		while(true){
 			if (s_mirror.length) s_registry.mirrorRegistry(s_mirror);
-			else s_registry.updatePackages();
```
WebFreak001: uhh you do realize this code is the same as the production code and you removed package updates with this?

wilzbach: Yes of course I do. That's why it's called experiment / WIP. It crashes sometimes when trying to update the registry while wget/curl is fetching. I just wanted to see how far I can get with a simple curl loop.

Anyhow, I can/will remove this once `--no-monitoring` is a thing (-> https://github.com/dlang/dub-registry/pull/310)

wilzbach:
```
7:46:36 AM: Build exceeded maximum allowed runtime
```

So I guess this would have to use a different approach, like
- generate responses directly from the registry
- use a different CI/build environment
- run request in parallel
- generate responses from the JSON file (though I think with https://github.com/dlang/dub-registry/pull/278 this wouldn't work anymore)

WebFreak001:
how about limiting packages to a select few? I doubt we will need all packages. And whenever packages are needed for a test they can get added (such as when people might want packages with different readme names, with different languages, with different metadata, different logos, different repo types, etc)

wilzbach:
This is not about previewing PRs, this is about having a static registry mirror on a CDN.

See: https://github.com/dlang/dub-registry/issues/309

WebFreak001:
oh ok nvm then, though for netlify PRs I think that would be a good idea

wilzbach:
> though for netlify PRs I think that would be a good idea

FYI: They are already limited because at the moment there's no way to reach a package if not via the main list view and dependencies of such a a package.
As all URLs with query parameters ("?") are ignored, the wget mirror won't do any pagination.

So if you go to https://dub.pm at the moment (that's the static snapshot = to what you see on the previews), only ~ 20 packages are reachable. Maybe a bit more with their dependencies.

MartinNowak:
Not sure that's the best move, https://github.com/dlang/dub-registry/issues/309#issuecomment-384923697.
It would also be a big future design liability.

PR #318, opened 2018-04-07
Replace assertion by enforcement to avoid crashing the registry.

s-ludwig:
Caused by a representation change of absolute paths between vibe-d:core and vibe-core.

Fixes dlang/dub#1441

### DIFF
--- a/source/dubregistry/dbcontroller.d
+++ b/source/dubregistry/dbcontroller.d
@@ -495,7 +495,7 @@ struct DbRepository {
 		auto path = url.path.bySegment;
 		if (path.empty)
 			throw new Exception("Invalid Repository URL (no path)");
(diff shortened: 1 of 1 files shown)


WebFreak001:
lol wtf how can a url not have an absolute path?

s-ludwig:
No, the representation of an absolute path is simply different between the old and the new implementation. The old implementation had an "absolute" flag for the path and then started with the first path segment after the initial "/". The new one in vibe-core is fully represented by the list of path segments, so it has to start with an empty segment to denote an absolute path.

At the time of writing the new implementation I didn't see a way to make this a non-breaking change, but what could be done would be to change `.bySegment` in the old implementation to explicitly prepend an empty segment if the absolute flag is set.

PR #324, opened 2018-04-20
Check array length to avoid range violation errors

### LINKED ISSUE #323: Crash of registry
made a typo when searching for the packages json url, found out that `https://code.dlang.org/packages/.json` kills the registry

edit: made a typo in the word typo

s-ludwig:
Fixes #323.


WebFreak001:
this would be an optimal use-case for findSplit here

PR #329, opened 2018-04-27
Fix mixup of "bitbucket" and "gitlab" repository kinds.

### LINKED ISSUE #327: Mixed up BitBucket and GitLab
Whenever I try to add the package from GitLab, it is interpreted as BitBucket and vice versa.

It seems like a bug is in this file: https://github.com/dlang/dub-registry/blob/796b01371e7702c67bd6ff2d5da1b9933fb32225/source/dubregistry/dbcontroller.d

s-ludwig:
Fixes #327.

### DIFF
--- a/source/dubregistry/dbcontroller.d
+++ b/source/dubregistry/dbcontroller.d
@@ -479,16 +479,17 @@ struct DbRepository {
 	string owner;
 	string project;
 
-	void parseURL(URL url) {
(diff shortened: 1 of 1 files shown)


s-ludwig:
Hm, auto-merge appears to be broken, it didn't work for any of the vibe.d repositories either.

PR #331, opened 2018-05-05
Make logo generation work for libevent.

s-ludwig:
The libevent driver doesn't support task joining across threads, so we are using std.concurrency to pass the results.


source/dubregistry/internal/utils.d
```diff
@@ -53,7 +56,7 @@ private struct LogoGenerateResponse
 }
 
 // need to return * here because stack returned values get destroyed for some reason...
```
WebFreak001: now this comment is a historical piece of software which will confuse any future developer. Don't fix what isn't broken doesn't really apply here ¯\\\_(ツ)\_/¯

PR #332, opened 2018-05-06
Fix #312 - allow to manually configure the documentation url

### LINKED ISSUE #312: Make it possible to either turn off the documentation tab or point it elsewhere
It looks like a "Documentation" tab was recently added to the projects on code.dlang.org, and code.dlang.org gives no control whatsoever over where it points to. It looks like it currently points to dpldocs.info, which in the case of dxml looks terrible because I actually went to the effort of properly defining link macros and the like in the documentation, which then of course aren't recognized by the tool being used. As such, I don't want there to be a link to http://dxml.dpldocs.info/dxml.html on the [code.dlang.org page for dxml](https://code.dlang.org/packages/dxml). I already have a link to the properly generated documentation in the readme that gets displayed. As such, I would like to either be able to get rid of the Documentation tab or point it to the proper documentation. The dpldocs bit may be great for a project that didn't bother to go to the effort of putting together nice documentation, but any project that actually really uses ddoc is going to look terrible with it. And now, because that Documentation tab is on the top in big letters, folks are more likely to click on it and see the badly generated documentation instead of the official documentation.

So, _please_ make it possible to at least configure what that Documentation tab points to.

wilzbach:
Allows the user to set the documentation url. By default (aka empty string) it will point to dpldocs.info.
The URL is only verified on a very naive level, but if the user wants to put stupid things in there, it will probably be hard to prevent him from doing so anyhow.

image

image

with the html pattern verification disabled:

image

CC @WebFreak001

### DIFF
--- a/source/dubregistry/dbcontroller.d
+++ b/source/dubregistry/dbcontroller.d
@@ -223,6 +223,11 @@ class DbController {
 		m_packages.update(["name": packname], update);
 	}
 
+	void setDocumentationURL(string packname, string documentationURL)
+	{
+		m_packages.update(["name": packname], ["$set": ["documentationURL": documentationURL]]);
+	}
+
 	bdata_t getPackageLogo(string packname, out bdata_t rev)
 	{
 		auto bpack = m_packages.findOne(["name": packname], ["logo": 1]);
@@ -472,6 +477,7 @@ struct DbPackage {
 	string[] categories;
 	long updateCounter = 0; // used to implement lockless read-modify-write cycles
 	@optional BsonObjectID logo; // reference to m_files
+	@optional string documentationURL;
 }
 
 struct DbRepository {
--- a/source/dubregistry/web.d
+++ b/source/dubregistry/web.d
@@ -743,7 +743,7 @@ class DubRegistryFullWebFrontend : DubRegistryWebFrontend {
 	}
 
 	@auth @path("/my_packages/:packname/set_documentation_url") @errorDisplay!getMyPackagesPackage
-	void postSetLogo(scope HTTPServerRequest request, string documentation_url, string _packname, User _user)
+	void postSetDocumentationURL(scope HTTPServerRequest request, string documentation_url, string _packname, User _user)
 	{
 		enforceUserPackage(_user, _packname);
 		auto isValidURL = documentation_url.empty || documentation_url.startsWith("http://", "https://");
--- a/source/dubregistry/web.d
+++ b/source/dubregistry/web.d
@@ -308,7 +308,17 @@ class DubRegistryWebFrontend {
 			//auto sampleURLs = ["test1", "test2"]; /* TODO: actually make this array exist and embed samples generated from repository */
 			string[] sampleURLs;
 			auto activeTab = req.query.get("tab", "info");
-			render!("view_package.dt", packageName, user, packinfo, versionInfo, readmeContents, sampleURLs, urlFilter, registry, activeTab);
+			render!("view_package.dt",
+					packageName,
+					user,
+					packinfo,
+					versionInfo,
+					readmeContents,
+					sampleURLs,
+					urlFilter,
+					registry,
+					activeTab,
+			);
 		}
 	}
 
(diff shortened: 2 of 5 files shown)


source/dubregistry/dbcontroller.d
```diff
@@ -472,6 +477,7 @@ struct DbPackage {
 	string[] categories;
 	long updateCounter = 0; // used to implement lockless read-modify-write cycles
 	@optional BsonObjectID logo; // reference to m_files
+	@optional string documentation_url;
```
s-ludwig: Naming convention -> `documentationURL`

wilzbach: Cool. I was actually unsure about this as I saw many places which still used the underscore convention.

views/my_packages.package.dt
```diff
@@ -41,6 +41,11 @@ block body
 		form(method="POST", enctype="multipart/form-data", action="#{req.rootDir}my_packages/#{packageName}/delete_logo")
 			button.danger(type="submit") Delete
 
+		form(method="POST", action="#{req.rootDir}my_packages/#{packageName}/set_documentation_url")
+			h3 Documentation URL
+			input(name="documentation_url", value="#{pack[`documentation_url`].get!string}", pattern="https?://.+")
```
s-ludwig: Missed one name check here (the second "documentation_url").

views/view_package.dt
```diff
@@ -99,7 +99,9 @@ block body
 				a.tab(class=activeTab == "info" ? "active" : "", href="#{req.rootDir}packages/#{normalizedPackageName}?tab=info") Info
 			- foreach (i, sample; sampleURLs)
 				a.tab(class=activeTab == "sample_"~i.to!string ? "active" : "", href="#{req.rootDir}packages/#{normalizedPackageName}?tab=sample_#{i}")= sample
-			a.tab.external(href="http://#{normalizedPackageName}.dpldocs.info/") Documentation
+			- auto doc_url = packinfo.info["documentation_url"].get!string;
```
s-ludwig: And this one.

views/my_packages.package.dt
wilzbach: Ah forgot to add `.dt` to the replace command and it still compiled because it's Json :/

source/dubregistry/web.d
```diff
@@ -732,6 +742,16 @@ class DubRegistryFullWebFrontend : DubRegistryWebFrontend {
 		redirect("/my_packages/"~_packname);
 	}
 
+	@auth @path("/my_packages/:packname/set_documentationURL") @errorDisplay!getMyPackagesPackage
```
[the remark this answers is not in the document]
wilzbach: Dumb auto-replace sorry. Good catch!

views/view_package.dt
```diff
@@ -99,7 +99,9 @@ block body
...
+			- doc_url = !doc_url.empty ? doc_url : "http://"~normalizedPackageName~".dpldocs.info";
```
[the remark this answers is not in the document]
wilzbach: Changed to https. Not really sure about the .empty vs .length part. AFAICT `.empty` is the idiomatic way to check for emptiness

wilzbach:
> imo it's about time we move and automate all these fields. I really dislike adding setters for everything here. Infos should be able to be versionable (multiple different documentation versions), they should be better supported by the API allowing for partial requests and they shouldn't all need to be @optional like that. (Maybe we could look into using https://github.com/WebFreak001/MongoSchemaD, which was designed to make DB <-> struct management a lot easier, though it seems like stuff has changed in the mongodb driver and on vibe.d stuff and it uses a custom serializer so that all needs some change)

Yeah I would like to see sth. like that too. I needed to touch five places in the backend for adding the documentation_url support. Ideally it should be one or two.
Anyhow, this would be an overkill for this PR.

> What bothers me especially is that in the diet templates we operate on a bson object containing this struct, making this struct basically useless and type-unsafe.

Yep that bothers me too, but I didn't want to change it in this PR.

s-ludwig:
>> What bothers me especially is that in the diet templates we operate on a bson object containing this struct, making this struct basically useless and type-unsafe.
>
> Yep that bothers me too, but I didn't want to change it in this PR.

Just to be precise (and yes, the current state is rather awful), it is operating on a JSON object that is generated from the BSON object and is even augmented with additional fields. But in fact, the `dubregistry.registry.PackageInfo` struct was already the first step in getting rid of this, it just needs to be extended appropriately.

Converting between `DbPackageInfo` and `PackageInfo` could then also be automated, but there is some merit in keeping both separate, to still have an abstraction layer between runtime representation and database representation. Especially when we hopefully change the target database type ad some point, or add support for alternative database types, that may turn out to be important to have.

But yeah, this is definitely nothing to be done in the context of this PR.

PR #336, opened 2018-05-07
Fix JSON typo in README.md

WebFreak001:
just caught this only after it has already been merged


WebFreak001:
how about adding `Typo` labels which are basically trivial + auto merge?

PR #337, opened 2018-05-07
[trivial] Avoid division by zero when computing the download score.

s-ludwig:
Happened for packages with zero downloads.


MartinNowak:
> Happened for packages with zero downloads.

Not quite @s-ludwig @wilzbach, happens when you don't have download infos. The package downloads are normalized to average downloads of all packages, so that the score automatically scales relative to the rest of the pack.
So in practice this is never zero, but when starting with an empty db it is.

PR #340, opened 2018-05-18
Use the full package count for pagination

### LINKED ISSUE #339: Just one page since last update ?
see here for example: at the bottom: https://code.dlang.org/?sort=added&category=&skip=0&limit=10,
where are 2 3 4 5 6 7 8 etc ?

s-ludwig:
The number of items on the first page got erroneously used as the total number of packages, thus always resulting in a single result page.

This got broken with the recent front page update: #303 

Fixes #339

### DIFF
--- a/source/dubregistry/web.d
+++ b/source/dubregistry/web.d
@@ -86,6 +86,7 @@ class DubRegistryWebFrontend {
 		} else {
 			packages = m_packages.dup;
 		}
+		auto pcount = packages.length;
 
 		// sort by date of last version
 		SysTime getDate(in ref CachedPackageSlot p) {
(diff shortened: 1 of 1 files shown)


source/dubregistry/web.d
```diff
@@ -116,7 +117,7 @@ class DubRegistryWebFrontend {
 		foreach (p; m_registry.getPackageInfos(packages.map!(p => p.name).array))
 			infos[p.info["name"].opt!string] = p.info;
 
-		return packages.map!(p => Package(p.stats, infos.get(p.name, Json.init))).array;
+		return tuple(packages.map!(p => Package(p.stats, infos.get(p.name, Json.init))).array, pcount);
```
wilzbach: `tuple!("list", "count")` instead of using rather cryptic indexes?

BBasile:
Do you need to wait another week for this ?

s-ludwig:
Nope, could have been merged already.

Issue #342, opened 2018-06-08
GitLab API v3 is no longer supported, use v4 instead


WebFreak001:
Tried to import a repository gave this error on the dub registry:

Failed to read JSON from https://gitlab.com/api/v3/projects/WebFreak001%2Fbancho-irc/repository/branches?private_token=sjpcNjkSC-a4dFBNnzzD: Unexpected reply for 'https://gitlab.com/api/v3/projects/WebFreak001%2Fbancho-irc/repository/branches?private_token=sjpcNjkSC-a4dFBNnzzD': Gone

Member1221:
Having this issue trying to update my package to a gitlab mirror, myself.

WebFreak001:
fixed by #345

PR #345, opened 2018-06-11
Change GitLab API to version 4

Member1221:
(This pull request is work in progress and is not tested yet, I do not have the means of testing if this works currently. But will test soon.)

I expect there to be breaking API changes between version 3 and 4. But I'll test when i get home.

GitLab 11.0 deprecated use of API v3 fully, therefore GitLab support is currently broken. See https://github.com/dlang/dub-registry/issues/342

### DIFF
--- a/source/dubregistry/repositories/gitlab.d
+++ b/source/dubregistry/repositories/gitlab.d
@@ -97,8 +97,8 @@ class GitLabRepository : Repository {
 		import std.uri : encodeComponent;
 		assert(path.absolute, "Passed relative path to listFiles.");
 		auto penc = () @trusted { return encodeComponent(path.toString()[1..$]); } ();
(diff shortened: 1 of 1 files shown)


WebFreak001:
can you check all api occurences for if they are still supported like they are used from v3 to v4?

Member1221:
Just tested it in a local build, it seems to work! Though you need to give it all of the permissions in the gitlab private access token dashboard.

Member1221:
Cleaned up some related code.

wilzbach:
I think only @MartinNowak and @s-ludwig are able to do deployment of code.dlang.org :/

s-ludwig:
Has been deployed now.

WebFreak001:
@s-ludwig can you check #347 ?

PR #347, opened 2018-06-11
Gitlab improvements

WebFreak001:
from #342 I saw that there were several useless HTTP requests and now the black function also censors query parameters starting with private or secret (fixing that the gitlab access token is dumped to the user)

### DIFF
--- a/source/dubregistry/internal/utils.d
+++ b/source/dubregistry/internal/utils.d
@@ -29,10 +29,11 @@ URL black(URL url)
 			if (part.startsWith("secret", "private")) {
 				if (!replace)
 					replace = url.queryString.dup;
-				auto eq = replace.indexOf('=', i) + 1;
-				if (eq < i + part.length) { // only replace value if possible (key=value)
-					replace = replace[0 .. eq] ~ "***" ~ replace[i + part.length .. $];
-					i = eq + 3;
+				auto eq = replace.indexOf('=', i);
+				if (eq != -1 && eq + 1 < i + part.length) { // only replace value if possible (key=value)
+					// +1 to pass the '=' character
+					replace = replace[0 .. (eq + 1)] ~ "***" ~ replace[i + part.length .. $];
+					i = eq + 4;
 				} else { // otherwise replace the whole pair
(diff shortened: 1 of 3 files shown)


source/dubregistry/internal/utils.d
```diff
@@ -10,17 +10,41 @@ import vibe.inet.url;
...
+			} else {
+				i += part.length;
+			}
+			i++; // '&'
+		}
+		if (replace)
+			url.queryString = (() @trusted => cast(string) replace)();
+	}
 	return url;
 }
```
s-ludwig: Would be good to have a unit test with the common cases, as the manual index manipulation is rather easy to get wrong.

source/dubregistry/internal/utils.d
```diff
@@ -10,17 +10,41 @@ import vibe.inet.url;
...
 	if (url.password.length > 0) url.password = "***";
+	if (url.queryString.length > 0) {
+		size_t i;
+		char[] replace;
+		foreach (part; url.queryString.splitter('&')) {
+			if (part.startsWith("secret", "private")) {
+				if (!replace)
+					replace = url.queryString.dup;
+				auto eq = replace.indexOf('=', i) + 1;
+				if (eq < i + part.length) { // only replace value if possible (key=value)
```
s-ludwig: Should be `eq >= 0` - `indexOf` returns `-1` if not found.

source/dubregistry/internal/utils.d
WebFreak001: done!

WebFreak001:
please note that all gitlab tokens up until this is merged and deployed should be deemed compromised. I would suggest also putting a warning somewhere and a changelog entry.

s-ludwig:
I'll take the opportunity and tag a new version, which hasn't been done for ages now. There is no change log so far, but I'll add a note in the README. Fortunately the account linked to the private token of the official registry is used for nothing else, so the private token is really not worth much in the wrong hands.

WebFreak001:
well that notice should be there for people hosting alternative registries

PR #353, opened 2018-06-28
Add support for bitbucket authentication

rracariu:
Adds support for private repositories on bitbucket.

### DIFF
--- a/source/dubregistry/repositories/bitbucket.d
+++ b/source/dubregistry/repositories/bitbucket.d
@@ -22,26 +22,30 @@ class BitbucketRepository : Repository {
 	private {
 		string m_owner;
 		string m_project;
+		string m_authUser;
+		string m_authPassword;
 	}
 
-	static void register()
+	static void register(string user, string password)
 	{
 		Repository factory(DbRepository info) @safe {
-			return new BitbucketRepository(info.owner, info.project);
+			return new BitbucketRepository(info.owner, info.project, user, password);
 		}
 		addRepositoryFactory("bitbucket", &factory);
 	}
 
-	this(string owner, string project)
+	this(string owner, string project, string auth_user, string auth_password)
 	{
 		m_owner = owner;
 		m_project = project;
+		m_authUser = auth_user;
+		m_authPassword = auth_password;
 	}
 
 	RefInfo[] getTags()
 	{
 		Json tags;
-		try tags = readJson("https://api.bitbucket.org/1.0/repositories/"~m_owner~"/"~m_project~"/tags");
+		try tags = readJson(getAPIURLPrefix ~ "/1.0/repositories/"~m_owner~"/"~m_project~"/tags");
 		catch( Exception e ) { throw new Exception("Failed to get tags: "~e.msg); }
 		RefInfo[] ret;
 		foreach (string tagname, tag; tags.byKeyValue) {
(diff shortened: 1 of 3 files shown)


WebFreak001:
add the new fields to readme pls

rracariu:
done.

source/dubregistry/repositories/bitbucket.d
```diff
@@ -120,13 +124,21 @@ class BitbucketRepository : Repository {
 		if( ver.startsWith("~") ) ver = ver[1 .. $];
 		else ver = ver;
 		auto venc = () @trusted { return encodeComponent(ver); } ();
+		if (m_authUser.length) return "https://"~m_authUser~":"~m_authPassword~"@bitbucket.org/"~m_owner~"/"~m_project~"/get/"~venc~".zip";
```
WebFreak001: I think adding another endpoint here (`getContentURLPrefix` for example) would be better in case something changes or another function uses it.

rracariu: Yeah - but YAGNI says otherwise. This is just on function and the URL is contained.

rracariu: Moved the URL to a common place and re-used the common sufix when auth is on. Hope this is OK.

source/dubregistry/repositories/bitbucket.d
```diff
@@ -83,7 +87,7 @@ class BitbucketRepository : Repository {
 	RepositoryFile[] listFiles(string commit_sha, InetPath path)
 	{
 		assert(path.absolute, "Passed relative path to listFiles.");
-		auto url = "https://bitbucket.org/api/2.0/repositories/"~m_owner~"/"~m_project~"/src/"~commit_sha~path.toString()~"?pagelen=100";
+		auto url = getAPIURLPrefix ~ "/api/2.0/repositories/"~m_owner~"/"~m_project~"/src/"~commit_sha~path.toString()~"?pagelen=100";
```
WebFreak001: I just noticed the prefixes changed, in the old code sometimes api.bitbucket.org and sometimes bitbucket.org/api was used. Is this documented somewhere which one to use now?

rracariu:
I followed this:  https://developer.atlassian.com/bitbucket/api/2/reference/meta/uri-uuid

WebFreak001:
oh ok lol these function were written by me, probably a typo then. Good to have that normalized now then. Still I don't really like the way you are manipulating the url right now, otherwise it looks great.

Can you use the URL struct and change username and password on it instead and then do toString? Just to be safe (in case someone changes for example http/https there)

Or instead of the 8 do `"https://".length` so it's clear where it comes from

rracariu:
Opted for the `"https://".length`.

source/dubregistry/repositories/bitbucket.d
```diff
@@ -120,13 +124,22 @@ class BitbucketRepository : Repository {
...
 
 	void download(string ver, scope void delegate(scope InputStream) @safe del)
 	{
 		downloadCached(getDownloadUrl(ver), del);
 	}
+
+	private string getAPIURLPrefix()
+	{
+		import std.uri : encodeComponent;
+		if (m_authUser.length) return "https://"~m_authUser~":"~m_authPassword~"@api.bitbucket.org";
```
WebFreak001: I think you actually want to use encodeComponent in here and not just leave the import there. Also do that in the function above.

rracariu: OK. Might wanna check the github provider for the same stuff.

PR #354, opened 2018-06-28
Auto-deploy to Dockerhub

wilzbach:
CC @rracariu - this is based on your Docker image, I hope you're okay with using your work to set up such an automatic pipeline?

DockerHub:

https://hub.docker.com/r/dlangcommunity/dub-registry/

(will add explanation to the README once everything else is working)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,3 +20,30 @@ matrix:
         env: VIBED_SSL=botan
 
 script: dub test --override-config="vibe-d:tls/${VIBED_SSL}"
+
+jobs:
+  include:
+    - stage: dockerhub-stable
+      d: dmd
+      os: linux
+      script:
+        - echo "Deploying to DockerHub..."
+        - dub build --build=release
+        - docker login -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD"
+        - LATEST_TAG="$(git describe --abbrev=0 --tags)"
+        - docker build -t "dlangcommunity/dub-registry:${LATEST_TAG} ."
+        - docker push "dlangcommunity/dub-registry:${LATEST_TAG}"
+      on:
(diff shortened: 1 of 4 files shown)


rracariu:
For sure. Keep credits though as license says https://github.com/rracariu/docker/blob/master/LICENSE

wilzbach:
> Keep credits though as license says

Ehm I think we need to license this under Boost as this entire repo is under Boost and would complicate things (especially in this area where companies might want to get involved).
I just added credit headers in the files where I looked at yours before. Would this be enough or did you have something else in mind?

rracariu:
OK, I can change the license to  Boost license for that image.

I'm happy you took over this - just wanted a small credit for this 😄

As long more people are contributing I'm happy to transfer this over to dlang and delete my image.

wilzbach:
> Original image is re-licensed as BSL.

Thanks a lot!

> So, maybe it will be safe to include the binary build with DMD as the binary for this image, or keep the LDC debug one.

Ok, I switched to DMD for now.

> Let me know when you feel your work is done so I can remove my version of the image and have this one as the official one.

I think this should be enough for an initial version. I can only verify that the auto-deploy really works by merging this.
Also should issues arise it's a lot easier to fix them incrementally.

The auto-deploy is configured to always push to the `latest` tag on every commit, but push a separate tag if it's a git tagged build/commit. I know this isn't the ideal way to do things, but it matches the current release model of the dub-registry (i.e. deploy a new version directly from master whenever there's a new feature merged).

wilzbach:
Seems like it works like a charm:

https://travis-ci.org/dlang/dub-registry/jobs/398269077
https://hub.docker.com/r/dlangcommunity/dub-registry/tags

PR #361, opened 2018-07-07
Use the global databaseName for the Download Cache

wilzbach:
Without this, registering new packages fails.

### DIFF
--- a/source/app.d
+++ b/source/app.d
@@ -72,10 +72,10 @@ void main()
 
 	version (linux) {
 		logInfo("Enforcing certificate trust.");
-		HTTPClient.setTLSSetupCallback((ctx) {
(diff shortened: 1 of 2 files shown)


wilzbach:
So I'm getting "Failed to load trusted certificate file" locally and on Heroku at the moment. Disabling the custom `setTLSSetupCallback` works as a workaround, but this shouldn't be required.
Ideas?

wilzbach:
Ok going to merge this now, s.t. we can deploy things to Heroku (and it's also failing locally for me) and will open a new PR which reverts the second commit and we can figure out what's happening before the next release.

wilzbach:
-> https://github.com/dlang/dub-registry/pull/363

PR #364, opened 2018-07-11
Dub build type syntax

andre2007:
Documentation enhanced for new build type "syntax":
https://github.com/dlang/dub/pull/1514


views/package_format_sdl.dt
```diff
@@ -657,6 +657,9 @@ block body
 		tr
 			td unittest-cov
 			td <code>"unittests" "coverage" "debugMode" "debugInfo"</code>
+		tr
+			td syntax
+			td <code>["syntaxOnly"]</code>
```
ohdatboi: In SDL brackets aren't required, it should be just `td <code>"syntaxOnly"</code>` instead

andre2007: Thanks, fixed

wilzbach:
In case someone is wondering with the recent split-up into dub-docs, I didn't forget to move this PR/commit:

https://github.com/andralex/dub-docs/commit/2f64eb1b36f04a7580c40d74dfdfba585784ae91

PR #371, opened 2018-07-15
Update input style

gedaiu:
On some os(eg. gnome shell with dark gtk theme) the default background colours for input boxes is black.
screenshot from 2018-07-15 20-20-28

This is a proposal for fixing this issue.

### DIFF
--- a/public/styles/common.css
+++ b/public/styles/common.css
@@ -339,6 +339,12 @@ button.primary:hover, button.primary:focus {
 
 .clipboard-pkg-row input {
 	width: 280px;
+	background: #fff;
+	color: #000;
+	border: 1px solid #ccc;
+	border-radius: 3px;
+	padding: 4px;
+	margin: 2px;
 }
 
 /********************/


WebFreak001:
@wilzbach can you make the deployment at least contain one package for testing PRs like this? It's not really useful to have a deployment without packages

wilzbach:
> @wilzbach can you make the deployment at least contain one package for testing PRs like this?

Well, the problem here is to distinguish between (a) people who deploy their private registry instance to heroku via the deploy button, (b) the staging instance and (c) the preview app. Distinguishing between (b) and (c) is easy as we can simply do sth. like: "if heroku and the database is empty".
However, I don't think you get acces to the slugname nor the exposed domain via Heroku.
For example, here's what `env` yields:

```
BUILDPACK_URL=https://github.com/ddollar/heroku-buildpack-multi.git
TERM=xterm-256color
COLUMNS=88
DYNO=run.9860
PATH=/usr/local/bin:/usr/bin:/bin
PWD=/app
PS1=\[\033[01;34m\]\w\[\033[00m\] \[\033[01;32m\]$ \[\033[00m\]
LINES=73
SHLVL=1
HOME=/app
PORT=47565
MONGODB_URI=mongodb://heroku***
_=/usr/bin/printenv
```

Anyhow, for now I simply created an account:
https://dub-registry-staging-pr-371.herokuapp.com/packages/mir-optim

wilzbach:
Update: here's an attempt at generating dummy data for Heroku Preview Apps: https://github.com/dlang/dub-registry/pull/376

PR #376, opened 2018-07-19
Generate dummy data for Heroku preview apps

wilzbach:


app.json
```diff
@@ -20,5 +20,11 @@
     "web": {
       "quantity": 1
     }
+  },
+  "env": {
+    "GENERATE_DEFAULT_DATA": {
+      "description": "Whether the database should be pre-filled. Use '0' for no.",
+      "required": false
+    }
```
wilzbach: See also: https://devcenter.heroku.com/articles/github-integration-review-apps#inheriting-config-vars

source/app.d
```diff
@@ -55,6 +55,21 @@ shared static this()
...
+{
+	if (environment.get("GENERATE_DEFAULT_DATA", "0") == "1" &&
+		registry.getPackageDump().empty)
+	{
+		logInfo("'GENERATE_DEFAULT_DATA' is set and an empty database has been detected. Inserting dummy data.");
+		auto userId = userMan.registerUser("dummy@dummy.org", "dummyUser",
+			"Dummy User", "test1234");
+		DbRepository repo;
+		repo.parseURL(URL("https://github.com/libmir/mir-optim"));
+		registry.addPackage(repo, userId);
```
wilzbach: We can add a few more once this is working (though we probably can't add too much due to the restriction of the unauthorized GitHub API).

source/app.d
```diff
@@ -55,6 +55,21 @@ shared static this()
...
 
+// generate dummy data for e.g. Heroku's preview apps
+void defaultInit(UserManController userMan, DubRegistry registry)
+{
+	if (environment.get("GENERATE_DEFAULT_DATA", "0") == "1" &&
+		registry.getPackageDump().empty)
+	{
+		logInfo("'GENERATE_DEFAULT_DATA' is set and an empty database has been detected. Inserting dummy data.");
+		auto userId = userMan.registerUser("dummy@dummy.org", "dummyUser",
+			"Dummy User", "test1234");
```
wilzbach: I don't think there's anything to be gained by exposing the dummy password here.
The worst one can due is to register a few packages (or delete them), but the review app is deleted after five days (or a merge) anyhow.
In case this really becomes a problem, we can always use an auto-generated password here.

wilzbach:
Oh nice!
This is already working:

image

Issue #349, opened 2018-06-18
Internal server error for projects hosted via gitlab


Member1221:
I am not sure what causes this issue.

Log from command: `dub fetch -v ppc`
```
Using dub registry url 'https://code.dlang.org/'
Refreshing local packages (refresh existing: true)...
Looking for local package map at /var/lib/dub/packages/local-packages.json
Looking for local package map at /home/clipsey/.dub/packages/local-packages.json
Determined package version using GIT: serve-d 0.1.3+commit.7.ga6a358e
Determined package version using GIT: trans-game ~master
Fetching ppc 0.0.2-alpha.2...
Downloading from 'https://code.dlang.org/packages/ppc/0.0.2-alpha.2.zip'
Getting a release version failed: HTTP request returned status code 500 (Internal Server Error)
Retry with ~master...
Fetching ppc ~master...
Downloading from 'https://code.dlang.org/packages/ppc/~master.zip'
HTTP request returned status code 500 (Internal Server Error)
```

ohdatboi:
I couldn't reproduce this issue on my local machine, but internal server error indeed happens when using registry at code.dlang.org.

One more affected package: trial. Downloading https://code.dlang.org/packages/trial/0.7.0-rc.1.zip fails with:
```
$ curl https://code.dlang.org/packages/trial/0.7.0-rc.1.zip   
<html>
<head><title>502 Bad Gateway</title></head>
<body bgcolor="white">
<center><h1>502 Bad Gateway</h1></center>
<hr><center>nginx/1.6.2</center>
</body>
</html>
```

gedaiu:
I can confirm that my package crashes the registry... I wonder why... maybe because I recently moved it to gitlab?

gedaiu:
I moved my repo back to github... It looks like the gitlab integration is broken..

wilzbach:
FWIW I tried this locally with the d package and from my Arch and the docker container we should I get no errros and the zip is correctly returned.
Maybe someone who can reproduce this locally can provide a stack trace (e.g. with `gdb` or `registerMemoryError` in `source/app.d`).

the-Mitu:
I noticed that for me dub-registry crashes only when I build with release (or release-debug) mode.

jmdavis:
[...]

PR #378, opened 2018-07-20
Support for GitLab projects inside subgroups

### LINKED ISSUE #374: Gitlab subgroups not supported
The URLs for GitLab projects involving subgroups are considered invalid:

Example URL: https://gitlab.com/group/subgroup/project

This makes me impossible to register the package for such projects.

the-Mitu:
This pull requests allows to register packages for GitLab projects placed into subgroups.

Currently the problem is that URL validation throws an error, because it expects url like *service*[.com]//*owner*/*project*. However, GitLab's subgroups allow URLs like gitlab.com/*owner*/*subgroup*/*subsubgroup*/*project*.

This pull request renames **m_project** to **m_projectPath** in **GitLabRepository** and allows it to store the *subgroup*/*subsubgroup*/*project* as is with slashes - so the first part after the URL is put into **m_owner** as it was before, but the rest is joined and put into **m_projectPath** even if contains slashes.

The regex is adjusted as well so that multiple slashed parts are allowed.

This will fix #374.

### DIFF
--- a/source/dubregistry/dbcontroller.d
+++ b/source/dubregistry/dbcontroller.d
@@ -509,9 +509,13 @@ struct DbRepository {
 		path.popFront;
 		if (path.empty || path.front.name.empty)
 			throw new Exception("Invalid Repository URL (missing project)");
-		project = path.front.name;
+
+		if(kind == "gitlab")
+			project = path.map!"a.name".join("/");
+		else
+			project = path.front.name;
 		path.popFront;
-		if (!path.empty)
+		if (!path.empty && kind != "gitlab")
 			throw new Exception("Invalid Repository URL (got more than owner and project)");
 	}
 
@@ -523,6 +527,8 @@ struct DbRepository {
 		assert(r == DbRepository("bitbucket", "bar", "baz"));
 		r.parseURL(URL("https://gitlab.com/foo/bar"));
 		assert(r == DbRepository("gitlab", "foo", "bar"));
+		r.parseURL(URL("https://gitlab.com/group/subgroup/project"));
+		assert(r == DbRepository("gitlab", "group", "subgroup/project"));
 		assertThrown(r.parseURL(URL("ftp://github.com/foo/bar")));
 		assertThrown(r.parseURL(URL("ftp://github.com/foo/bar")));
 		assertThrown(r.parseURL(URL("http://github.com/foo/")));
--- a/source/dubregistry/dbcontroller.d
+++ b/source/dubregistry/dbcontroller.d
@@ -510,7 +510,7 @@ struct DbRepository {
 		if (path.empty || path.front.name.empty)
 			throw new Exception("Invalid Repository URL (missing project)");
 
-		if(kind == "gitlab")
+		if(kind == "gitlab")  // Allow any number of segments, as GitLab's subgroups can be nested
 			project = path.map!"a.name".join("/");
 		else
 			project = path.front.name;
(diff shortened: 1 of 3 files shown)


the-Mitu:
Unfortunately I have only private projects in subgroups. I've tested it locally on them and it worked but I expect you want to test it as well.

I've browsed the public GitLab groups and found some projects laying inside a subgroup that is publicly available. Example one: https://gitlab.com/ezmaxtech/ha/register

That one is actually not a D project, but maybe it is sufficient to see that the repo is accessed and an error appears just because there is are no dub projects in it. Alternatively a test group/subgroup has to be created.

Thanks for noting the #349 - however I actually need to set up my own instance for hosting non-public packages, so I'll hope I'm more lucky. :)

the-Mitu:
@wilzbach, Tried to test on staging - "Unknown repository type: gitlab". Haven't you forgotten to set the GitLab's server URL in settings.json?

I'd actually suggest it to be "gitlab.com" by default unless another one is specified - not unsupported unless specified - but this is something for another ticket.

wilzbach:
@the-Mitu 
- the staging server is very new (we set it up this week)
- there's no settings.json on Heroku, only environment variables

> this is something for another ticket.

https://github.com/dlang/dub-registry/pull/379
tl;dr: after #379 we should be able to test this here too.

> I've browsed the public GitLab groups and found some projects laying inside a subgroup that is publicly available. Example one: https://gitlab.com/ezmaxtech/ha/register

Hmm I think it needs a valid dub.json / dub.sdl
Anyhow, I created https://gitlab.com/dub-registry-test/a-subgroup/test (it's a fork of https://gitlab.com/dub-registry/test).

> however I actually need to set up my own instance for hosting non-public packages, so I'll hope I'm more lucky. 

We have a #dub channel on the D Slack where you might be able to find help if you aren't lucky.
Simply send me a ping if you want an invite.

wilzbach:
> after #379 we should be able to test this here too.

Merged and thus I rebased this PR. We should be able to test https://gitlab.com/dub-registry-test/a-subgroup/test in a few minutes.

views/my_packages.register.dt
```diff
@@ -15,13 +15,13 @@ block body
...
 			p.light
-				a.blind(href="#{req.rootDir}publish") Learn more 
+				a.blind(href="#{req.rootDir}publish") Learn more
 				| about publishing.
 
 		form(method="POST", action="#{req.rootDir}register_package")
 			p
 				label(for="project") Repository URL
-				input(type="text", name="url", value=url, pattern="^(https?://)?(www\.)?(github(\.com)?|gitlab(\.com)?|bitbucket(\.org)?)/[^/]+/[^/]+([?#.].*)?$", placeholder="github.com/user/repository")
+				input(type="text", name="url", value=url, pattern="^(https?:\/\/)?(www\.)?(github(\.com)?|gitlab(\.com)?|bitbucket(\.org)?)(/[^/]+)+/[^/]+([?#.].*)?$", placeholder="github.com/user/repository")
```
wilzbach: This allows invalid patterns for GitHub & Bitbucket, e.g. github.com/foo/bar/barrrr

source/dubregistry/dbcontroller.d
```diff
@@ -509,9 +509,13 @@ struct DbRepository {
...
 			throw new Exception("Invalid Repository URL (missing project)");
-		project = path.front.name;
+
+		if(kind == "gitlab")
+			project = path.map!"a.name".join("/");
+		else
+			project = path.front.name;
 		path.popFront;
-		if (!path.empty)
+		if (!path.empty && kind != "gitlab")
```
wilzbach: This would allow gitlab.com/foo/bar/foobar/fooobar/foo -  is this really possible?

edit: oh it looks like subgroups can be nested in subgroups

A comment about this in the source code would help future readers...

wilzbach:
>  We should be able to test https://gitlab.com/dub-registry-test/a-subgroup/test in a few minutes.

https://dub-registry-staging-pr-378.herokuapp.com/packages/gitlab-test-package :)

views/my_packages.register.dt
the-Mitu: Yes, it will. The right regex is:

    ^(https?:\/\/)?(www\.)?(((github(\.com)?|bitbucket(\.org)?)\/[^/]+\/[^/]+([?#.].*)?)|(gitlab(\.com)?(\/[^/]+)+\/[^/]+([?#.].*)?))$

I'll put it in the code tomorrow and commit the fix. In the meantime I would be grateful if anyone tested it once again to confirm its correctness, as it gets long now.

source/dubregistry/dbcontroller.d
the-Mitu: I confirm, GitLab's subgroups are nestable. I'll add the comment as well

the-Mitu: Comment added.

the-Mitu:
@wilzbach, could I get that invitation to Slack workspace? I'm probably affected by the error you mentioned - and also have a few questions on other topic as well.

wilzbach:
@the-Mitu Invitation is out.

PR #379, opened 2018-07-20
Environment fallback for config variables

wilzbach:
Allows to set the config variables optionally via environment variables (soft fallback)

This is useful for e.g. the Heroku Apps where the app is ephermal and all configuration is done via environment variables, which can be inherited automatically to new review instances.

### DIFF
--- a/source/app.d
+++ b/source/app.d
@@ -102,6 +102,16 @@ void main()
 	bool noMonitoring;
 	setLogFile("log.txt", LogLevel.diagnostic);
 
+    version (linux)
+    {
+    	// register memory error handler on heroku
+    	if ("DYNO" in environment)
+    	{
+			import etc.linux.memoryerror : registerMemoryErrorHandler;
+        	registerMemoryErrorHandler();
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


source/app.d
```diff
@@ -70,11 +70,48 @@ void defaultInit(UserManController userMan, DubRegistry registry)
...
 	bool noMonitoring;
 	setLogFile("log.txt", LogLevel.diagnostic);
 
+    version (linux)
+    {
+    	// register memory error handler on heroku
+    	if ("DYNO" in environment)
+    	{
+			import etc.linux.memoryerror : registerMemoryErrorHandler;
+        	registerMemoryErrorHandler();
```
wilzbach: This was very helpful during debugging. It might even make sense to always activate this, but for now I only enabled this for Heroku (there's no way to attach `gdb` to your instance or get a coredump anyhow).

wilzbach:
I added GitHub & GitLab configuration environments variables to Heroku and it works nicely (they are automatically inherited to the Review App):
https://dub-registry-staging-pr-379.herokuapp.com/packages/gitlab-test-package

PR #391, opened 2018-11-17
Use new `dub add` command instead of copying json/sdl

Flying-Toast:
**This PR shouldn't be merged until a new dub version is released with dlang/dub#1587.**

dlang/dub#1587 creates an `add` command that adds a dependency, so instead of copy/pasting `"vibe-d": "~>0.8.4"` you just have to run `dub add vibe-d`.

### DIFF
--- a/views/view_package.dt
+++ b/views/view_package.dt
@@ -64,20 +64,13 @@ block body
 
 		p= versionInfo["description"].opt!string
 		br
-		p To use this package, put the following dependency into your project's dependencies section:
+		p To use this package, run the following command in your project's root directory:
 		.clipboard-pkg
 			- string expr = versionInfo["version"].get!string;
 			- if( !expr.startsWith("~") ) expr = "~>" ~ expr;
 			.clipboard-pkg-row
-				.clipboard-pkg-title dub.json
-				input(id="package-clipboard-json",value='"#{packageName}": "#{expr}"')
-				button(class="btn-clipboard",data-clipboard-target="#package-clipboard-json")
-					img(src="/images/clippy.svg", width=13, alt="Copy to clipboard")
-
-			.clipboard-pkg-row
-				.clipboard-pkg-title dub.sdl
-				input(id="package-clipboard-sdl",value='dependency "#{packageName}" version="#{expr}"')
-				button(class="btn-clipboard",data-clipboard-target="#package-clipboard-sdl")
+				input(id="package-clipboard",value='dub add #{packageName}')
+				button(class="btn-clipboard",data-clipboard-target="#package-clipboard")
 					img(src="/images/clippy.svg", width=13, alt="Copy to clipboard")
 
 		- if (versionInfo["subPackages"].opt!(Json[]).length)
--- a/public/styles/common.css
+++ b/public/styles/common.css
@@ -331,12 +331,6 @@ button.primary:hover, button.primary:focus {
 	padding-top: 5px;
 }
 
-.clipboard-pkg-title {
-	font-weight: bold;
-	display: inline-block;
-	width: 60px;
-}
-
 .clipboard-pkg-row input {
 	width: 280px;
 	background: #fff;


thewilsonator:
Please ping me when that happens, I'm likely to forget.

MartinNowak:
Mmh dub add doesn't support specific versions (e.g. see [A](http://code.dlang.org/packages/vibe-d/0.8.4) and [B](http://code.dlang.org/packages/vibe-d/0.7.30)).
Most of the time only the latest is needed, but would be nice if that was supported, e.g. as `dub add vibe-d='~>0.8.4' dpq2='~>1.0.5'`. 
Shouldn't be too hard to still add version support to dub's add command before we release 2.084.0.

MartinNowak:
See https://github.com/dlang/dub/pull/1626 for version-spec support in `dub add` and `dub init`.

Flying-Toast:
@thewilsonator 2.084.0 was released.
Do you think it would be better to wait a while longer, to give people time to update?

wilzbach:
Let's wait at least until LDC's next release with 2.084.

Flying-Toast:
@wilzbach LDC 1.14.0 was released yesterday :tada:

Flying-Toast:
Oops. Why is the "Close and comment" button right next to the "Comment" button?! :laughing:

views/view_package.dt
```diff
@@ -64,20 +64,13 @@ block body
...
-				.clipboard-pkg-title dub.json
-				input(id="package-clipboard-json",value='"#{packageName}": "#{expr}"')
-				button(class="btn-clipboard",data-clipboard-target="#package-clipboard-json")
-					img(src="/images/clippy.svg", width=13, alt="Copy to clipboard")
-
-			.clipboard-pkg-row
-				.clipboard-pkg-title dub.sdl
-				input(id="package-clipboard-sdl",value='dependency "#{packageName}" version="#{expr}"')
-				button(class="btn-clipboard",data-clipboard-target="#package-clipboard-sdl")
+				input(id="package-clipboard",value='dub add #{packageName}')
```
wilzbach: Shouldn't we include the version here?

wilzbach:
> @wilzbach LDC 1.14.0 was released yesterday

Hmm, but it takes quite some time until at least the majority of users (and distro) have upgraded to it, no?
The upcoming disco release of Ubuntu (19.04) will most likely ship with LDC 1.12 (and Dub 1.12):

https://packages.ubuntu.com/disco/ldc
https://packages.ubuntu.com/disco/dub

> Do you think it would be better to wait a while longer, to give people time to update?

Maybe as a compromise we can show both instructions for this intermediate period?

ZombineDev:
As there are plenty of ways to obtain the latest version of dub (install.sh, snapd, tar.gz download, etc.) I see no reason not to show instructions for the latest and greatest version.
If we really want, we could display an ℹ️ info tooltip reminding users that they need a recent version of dub.

views/view_package.dt
Flying-Toast: `dub add` will query the registry for the latest version. `dub add packageName=1.2.3` would do the same thing as just `dub add packageName`

wilzbach: Yes, but by supplying the exact version we can make sure the user gets an error if there's an issue with the registry mirror he's using. Also currently we advertise the current version:

image

I had a quick look around. Cargo does it the same way, but PIP and NPM also don't use the exact version. Other opinions?

wilzbach:
CC @s-ludwig and @MartinNowak. What do you think about this? Should we wait longer or release this now?

wilzbach:
No one objected, so let's do this with the next release of the registry.

Flying-Toast:
So should we also merge dlang/dub-docs#13 for consistency?

0xEAB:
This should really have been `additionally` instead of `instead` :roll_eyes:

Issue #93, opened 2015-02-04
Search issue


xaqq:
Hello,

When searching for `zmq` on the dub package registry (http://code.dlang.org/search?q=zmq) only one result is returned: `zmq-d`.

There is however another package, `zmqd` (the one I was looking for) that exists (http://code.dlang.org/packages/zmqd), but the search didn't return it.
I believe this to be an issue.

Thanks

MartinNowak:
We should setup an ElasticSearch instance to solve these issues.
This would also allow to search in the readme, fix spelling and provide suggestions.

MartinNowak:
Related https://github.com/D-Programming-Language/dub-registry/issues/86.

MartinNowak:
Just found https://github.com/intrica/elasticsearch-d, didn't find it using the search though :(.
Maybe @intrica can help us setting up an ElasticSearch server for the dub-registry?

s-ludwig:
In the meantime (and as long as we are as small as now), we could also go for the brute force route and simply scan through all packages linearly. I could quickly add something using `levenshteinDistance` and see how fast it is.

s-ludwig:
A simple version is running now, seems to be fast enough for the moment. It still needs some tweaking, though (e.g. any three letter sequence will match any three letter word).

s-ludwig:
Should be relatively usable now. It would be nice to also match parts of a word, so that for example "elastic search" matches "elasticsearch".

MartinNowak:
> I could quickly add something using levenshteinDistance and see how fast it is.

Guess you don't know this?
https://github.com/D-Programming-Language/dub/blob/ad9b43d6a150b53872db73c161e745c000cf8577/source/dub/internal/utils.d#L246
https://github.com/D-Programming-Language/dub/pull/453

MartinNowak:
> A simple version is running now, seems to be fast enough for the moment. It still needs some tweaking, though (e.g. any three letter sequence will match any three letter word).

And there is an easy opportunity to make the function much faster.
Also there is a more space efficient variant of the algorithm.
http://en.wikipedia.org/wiki/Levenshtein_distance#Iterative_with_two_matrix_rows
[Issue 13834 – make levenshteinDistance @nogc](https://issues.dlang.org/show_bug.cgi?id=13834)

s-ludwig:
> Guess you don't know this?

[...]

PR #399, opened 2018-12-24
improve FTS index

MartinNowak:
- see #93
- drop readme index (too much noise for MongoDB's simple search)
- search package name, categories, description, and authors
  (with decreasing weights)
- fix unwanted indexing of other text-fields (`$**` key)
- fix indexing of authors

### DIFF
--- a/source/dubregistry/dbcontroller.d
+++ b/source/dubregistry/dbcontroller.d
@@ -84,19 +84,24 @@ class DbController {
 
 		// drop old text index versions
 		db.runCommand(["dropIndexes": "packages", "index": "packages_full_text_search_index"]);
+		db.runCommand(["dropIndexes": "packages", "index": "packages_full_text_search_index_v2"]);
 
 		// add current text index
-		Bson[string] fts;
-		fts["key"] = ["$**": Bson("text")];
-		fts["name"] = "packages_full_text_search_index_v2";
-		fts["weights"] = [
-			"name": Bson(4),
-			"categories": Bson(3),
-			"versions.info.description" : Bson(3),
-			"versions.info.homepage" : Bson(1),
-			"versions.info.author" : Bson(1),
-			"versions.readme" : Bson(2),
+		immutable keyWeights = [
+			"name": 8,
+			"categories": 4,
+			"versions.info.description": 2,
+			"versions.info.authors": 1
 		];
+		Bson[string] fts;
+		fts["key"] = Bson.emptyObject;
+		fts["weights"] = Bson.emptyObject;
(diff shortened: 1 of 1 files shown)


source/dubregistry/dbcontroller.d
```diff
@@ -84,19 +84,24 @@ class DbController {
...
-		fts["weights"] = [
-			"name": Bson(4),
-			"categories": Bson(3),
-			"versions.info.description" : Bson(3),
-			"versions.info.homepage" : Bson(1),
-			"versions.info.author" : Bson(1),
-			"versions.readme" : Bson(2),
+		immutable keyWeights = [
+			"name": 8,
+			"categories": 4,
```
WebFreak001: should categories really have a text index? After all it just stores the category ids which just happen to be human-readable right? I think it would decrease search quality for generic terms, the category should be a separate UI dropdown menu or something to filter by before searching.

wilzbach: The idea is that if you enter "database" all packages in this category will be shown too (even if their name doesn't match).

WebFreak001: what about categories like std_aspirant, data_structures, optimized_mem, etc.

my point is that searching for the identifier of the category just happens to make sense for some existing categories, but will not work with other category labels.

It would be much better if we had a separate category search, or embedded it in the input field like `category:database` with autocompletion for it

source/dubregistry/dbcontroller.d
```diff
@@ -325,14 +330,22 @@ class DbController {
...
+
+		// normalize textScore to same scale as package score
+		immutable minMaxTS = pkgs.map!(p => p.textScore).fold!(min, max)(0.0f, 0.0f);
+		immutable scale = (DbPackageStats.maxScore - DbPackageStats.minScore) / (minMaxTS[1] - minMaxTS[0]);
+		foreach (ref pkg; pkgs)
+			pkg.textScore = (pkg.textScore - minMaxTS[0]) * scale + DbPackageStats.minScore;
+
+		// sort found packages by weighted textScore and package score
+		return pkgs
+			.sort!((a, b) => a.stats.score + 2 * a.textScore > b.stats.score + 2 * b.textScore)
```
wilzbach: This is a bit weird as ideally this should be handled by the database, but I'm guessing MongoDB doesn't support this :/

source/dubregistry/dbcontroller.d
wilzbach: I think the plan was to get rid of these categories and replace them with tags anyhow as categories have a rather bad UX.

PR #401, opened 2018-12-31
Redirect downloads to GitHub Releases

wilzbach:
We have been using them for quite a while now and the dlang-bot uploads are pretty stable.
There has been quite some confusions about the Dub downloads being outdated, so this should stop this.


s-ludwig:
We should set up a Windows build, though. My (currently defunct) CI worker just executed "build.cmd" and "scripts\win-installer\make_installer.cmd" (with NSIS having to be installed).

Geod24:
Came here to report an issue: `Download` currently references DUB 1.11 (!).
Would be great to have this in.

s-ludwig:
Can't really see a reasonable argument against this, except for the merge conflict.

wilzbach:
Great. I resolved the conflict (:

Re: windows builds. Yeah, we might be able to setup Azure builds. This would also help the normal dub binaries as only when build with Visual Studio they don't have a dependency on msvcr100.dll.

(see e.g. https://issues.dlang.org/show_bug.cgi?id=19760)

Issue #406, opened 2019-02-12
.travis-ci.yml is failing due to botan deprecation errors


ljmf00:
The current config of travis-ci is failling because https://github.com/etcimon/botan package is basically dead and outdated.

Is `botan` really worth it? `dub-registry` seems using `openssl` on production, so at least it works.

wilzbach:
Yeah, botan looks to be unmaintained (see also: https://github.com/etcimon/botan/issues/49).

Thus, probably the best action is to remove those two jobs (or at least allow them to fail).

ljmf00:
Fixes with #407 .

PR #407, opened 2019-02-13
Allow botan to fail on travis-ci

ljmf00:
Workarround for issue #406 . As referred, `botan` looks unmaintained and `openssl` seems used in production.


wilzbach:
Thanks!

PR #410, opened 2019-02-16
New tabbed edit package UI

WebFreak001:
This makes it easier to extend the package configuration and add more settings without cluttering the UI

Tip for reviewing: disable whitespace in diff options

using tabs (if JS is enabled)

**Screenshots:**

<details>
<summary>Desktop view</summary>

package view general tab
package view categories tab
package view repository tab
package view advanced tab

</details>

---

<details>
<summary>Mobile view</summary>

mobile top view of general
mobile bottom view of general

</details>

---

<details>
<summary>no-JS view (kind of like before, but with buttons at the top to jump around)</summary>

no-JS dub package view

</details>

---

also package view looks slighly different now to better highlight which tab is selected:
package documentation tab now slightly further down


wilzbach:
A bit unrelated, but maybe we should hide warnings >= 5 with e.g. `<details>`?
I experimented with DCD as test package and got this:

image

WebFreak001:
imo there should be a way to dismiss warnings because there is no way to fix them anyway

views/my_packages.package.dt
```diff
@@ -15,151 +15,136 @@ block body
...
+			section.contents
+				p
+					- auto stats = registry.getPackageStats(packageName).downloads;
+					| #[strong= stats.daily.to!string] downloads today
+						br
+					| #[strong= stats.weekly.to!string] downloads this week
+						br
+					| #[strong= stats.monthly.to!string] downloads this month
+						br
+					| #[strong= stats.total.to!string] downloads total
```
wilzbach: I guess this could be a separate tab "statistics"?

public/scripts/pkgadmin.js
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * @param {HTMLElement} subtabs
+ */
+function upgradeSubtabs(subtabs, openpage) {
+	if (openpage === undefined)
+		openpage = window.location.hash.substr(1);
+
+	var links = subtabs.querySelectorAll("a.tab");
+
+	if (openpage)
```
wilzbach: openpage vs. hasPage below -> be consistent in naming.

public/scripts/pkgadmin.js
```diff
@@ -0,0 +1,116 @@
...
+	{
+		var hasPage = false;
+		for (var i = 0; i < links.length; i++)
+			if (links[i].getAttribute("data-tab") == openpage)
+				hasPage = true;
+		if (!hasPage)
+			return;
+	}
```
wilzbach: Move into separate function

public/scripts/pkgadmin.js
```diff
@@ -0,0 +1,116 @@
...
+	var subtabs = document.querySelectorAll(".subtabsHeader");
+	for (var i = 0; i < subtabs.length; i++)
+		upgradeSubtabs(subtabs[i], openpage);
+}
+
+upgradeAllSubtabs("");
+if (window.location.hash.length)
+	upgradeAllSubtabs(window.location.hash.substr(1));
+
+window.onhashchange = function() { upgradeAllSubtabs(); };
```
wilzbach: better add a proper event listener here (otherwise this might be overwritten by other scripts)

public/scripts/pkgadmin.js
```diff
@@ -0,0 +1,116 @@
...
+
+function upgradeAllSubtabs(openpage) {
+	var subtabs = document.querySelectorAll(".subtabsHeader");
+	for (var i = 0; i < subtabs.length; i++)
+		upgradeSubtabs(subtabs[i], openpage);
+}
+
+upgradeAllSubtabs("");
+if (window.location.hash.length)
+	upgradeAllSubtabs(window.location.hash.substr(1));
```
wilzbach: Why not `if ... else`?

public/scripts/pkgadmin.js
```diff
@@ -0,0 +1,116 @@
...
+	subtab.style.display = show ? "" : "none";
+}
+
+function upgradeAllSubtabs(openpage) {
+	var subtabs = document.querySelectorAll(".subtabsHeader");
+	for (var i = 0; i < subtabs.length; i++)
+		upgradeSubtabs(subtabs[i], openpage);
+}
+
+upgradeAllSubtabs("");
```
wilzbach: Be consistent. The call below uses no arguments.

WebFreak001: passing empty string initializes with default page, passing with no args will put in the current hash

wilzbach:
Looking at it this is a very nice improvement :+1:, but I think we should try to make the "General" tab a bit prettier. It looks very unstructured at the moment:

image

Also, now that we have tabs I guess we could add "Badges" as another one and show sth. similar to

image

Maybe in a style similar to:

image

But I will open a separate issue for this.

views/my_packages.package.dt
WebFreak001: I quite like having the statistics directly when you open the package instead.

Maybe a more dedicated border and some more stats would make this better.

WebFreak001:
I did slight modifications making the general page look a bit better
new general page

wilzbach:
Have you tried moving the logo to the right side and inserting a border in-between?

wilzbach:
Alternatively if this still looks weird, maybe the logo can always be a separate tab (as its not too often updated)?

WebFreak001:
I think it's wasted space moving that into a separate tab (and like this it directly shows you an image of what you are editing), so I would definitely keep it in the general tab.

I will add a small separator line

WebFreak001:
anything else for this PR? It's fairly trivial to add new tabs (just a button and a new div with anchor like the others) so anything additional could also be added later. I think as for how this is right now, this is a fairly well working solution.

wilzbach:
(preview can be done with "View deployment" and the user `dummyUser` with `test1234` - this user is created freshly for each preview)

thewilsonator:
>@thewilsonator maybe?

I'm really not good with web stuff, @ZombineDev?

wilzbach:
Ping. Could anyone spare a second pairs of eyes here? 
Maybe create a thread in the NG?

ZombineDev:
I was not able to login with those credentials.

wilzbach:
Ah. The registry seems to auto-lowercase the username on save, but not when checking for the login. Try `dummyuser` or `dummy@dummy.org`

public/styles/common.css
```diff
@@ -647,6 +653,127 @@ div.packageInfo > .actions > a + p.detail {
 	}
 }
 
+/********************/
+/* PACKAGE ADMIN    */
+/********************/
+
+section.pkgconfig:after
+{
```
Flying-Toast: This file has inconsistent bracket styles.

PR #414, opened 2019-02-19
Add webhooks for instant package updates - Fix #412

### LINKED ISSUE #412: Update package webhook
there should be generic and github compatible webhooks for triggering a package update (so users can configure them to trigger on tag/branch creation)

A secret would be stored with every package to prevent abuse.

When #410 is merged it shouldn't be a problem to add instructions for it to the package management UI

WebFreak001:
- [ ] Depends on #410 (contains that PR already for having new location in package edit page)

I put a http-update.md during development in the root folder just to collect some thoughts how I wanted to implement it, that file will need to be moved to some proper API documentation page (which I don't think we have yet)


http-update.md
```diff
@@ -0,0 +1,24 @@
...
+## `POST /api/packages/:packageName/update`
+
+Queues an update for the specified package.
+
+Query params:
+`secret`: string (optional) provide the secret as query
+`header`: string (optional) provide which header is used to check the secret (must start with X-)
+
+Body params: (application/x-www-form-urlencoded, multipart/form-data or application/json)
+`secret`: string (optional) provide the secret as body param
```
wilzbach: Hmm, what's the use case for this? Isn't this prone to be over-used/falsely used?
(I'm referring to the API method itself)

views/my_packages.register.dt
```diff
@@ -16,7 +16,7 @@ block body
 		h1 Add new package
 			p.light
 				a.blind(href="#{req.rootDir}publish") Learn more
-				| about publishing.
+				|  about publishing.
```
wilzbach: Unrelated

http-update.md
```diff
@@ -0,0 +1,24 @@
...
+
+## `POST /api/packages/:packageName/update/github`
+
+Queues an update for the specified package. Compatible with GitHub webhooks and only triggers on `create` events. Must pass secret as query param and not in GitHub webhook settings.
```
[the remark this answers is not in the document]
WebFreak001: I wanted to implement github properly at first, but vibe.d with rest interface doesn't give me a way to get the full request body as string, which I need for the SHA1 HMAC verification with the GitHub secret field (at least I didn't find any way)

views/my_packages.register.dt
WebFreak001: this comes from the other PR, we should get that one merged before doing this PR. I just put it here for now for information and how the implementation will basically look.

http-update.md
WebFreak001: I wanted to make sure it's easy to incoperate this into any service you might be using for webhooks, so I'm offering to make the API callable with a secret per query string, body param or per HTTP Header.

http-update.md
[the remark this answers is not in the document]
WebFreak001: Also I don't think any of the other dub API currently has rate limiting yet. This will only queue a recheck and is basically the same as spamming the "check for updates" button on your project page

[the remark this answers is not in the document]
wilzbach: > I wanted to implement github properly at first,

Don't reinvent the wheel -> https://github.com/dlang/dlang-bot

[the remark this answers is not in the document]
wilzbach: > basically the same as spamming the "check for updates" button on your project page

Fair point, but by providing hooks we make it a lot easier for people to spam / run us into the rate-limits of GitHub.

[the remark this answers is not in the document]
WebFreak001: > > I wanted to implement github properly at first,
> 
> Don't reinvent the wheel -> [dlang/dlang-bot](https://github.com/dlang/dlang-bot)

I more accurately meant I wanted to support the "Secret" field on GitHub which sends a Signature of the body text as SHA1 HMAC with the Secret as password, but with vibe.d the rest interface implementation already ate the body so I couldn't access the raw string from code to hash. For hashing the stdlib provides everything needed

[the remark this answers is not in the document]
s-ludwig: Regarding the rate limiting, if we have web hook support, I would propose to stretch the polling over a much larger period than now (e.g. 6 hours vs. 30 minutes), so that it merely acts as a fallback in case the web hooks fail. This should reduce the risk of running into the rate limit considerable when compared to the current state.

Non-webhook repositories could be kept at the 30 min. polling interval, of course.

Regarding the signature issue, I believe this should work using an `@before` parameter that reads the body, validates the signature, and then parses and returns the JSON.

source/dubregistry/registry.d
```diff
@@ -429,6 +431,38 @@ class DubRegistry {
...
+	}
+
+	void regenPackageSecret(string pack_name)
+	{
+		import std.ascii : lowerHexDigits;
+		import std.random : uniform;
+
+		char[24] token;
+		foreach (ref c; token)
+			c = lowerHexDigits[uniform(0, $)];
```
WebFreak001, line 461: TODO: use secure random instead of uniform

s-ludwig:
I missed this PR at the time, but I think it would be fine to add it like this. It wouldn't completely "fix" the original issue due to its manual nature, but it would be nice to have as a convenience feature for those willing to do some configuration work.

Just getting a decent percentage of the packages being updated by push notifications would still require OAuth+PubSubHubub or a similar solution that works more or less seamlessly.

source/dubregistry/api.d
```diff
@@ -187,6 +245,18 @@ private:
 	}
 }
 
+private string updateSecretReader(scope HTTPServerRequest req, scope HTTPServerResponse res)
```
WebFreak001, line 248: this needs a better name

source/dubregistry/api.d
```diff
@@ -187,6 +245,18 @@ private:
...
+{
+	string header = req.query.get("header", "");
+	if (header.length)
+		return req.headers.get(header);
+
+	string ret = req.contentType == "application/json" ? req.json["secret"].get!string : req.form.get("secret", "");
```
WebFreak001, line 254: probably want to tryIndex the JSON

source/dubregistry/api.d
```diff
@@ -176,6 +188,52 @@ override {
...
 	}
+
+	@before!updateSecretReader("secret")
+	string postUpdate(string _name, string secret = "")
+	{
+		if (!secret.length)
+			return "No secret sent";
+
+		string expected = m_registry.getPackageSecret(_name);
+		if (!expected.length || secret != expected)
```
WebFreak001, line 199: use constant-time comparison

source/dubregistry/registry.d
```diff
@@ -429,6 +431,38 @@ class DubRegistry {
...
+
+		m_db.setPackageSecret(pack_name, token[].idup);
```
WebFreak001, line 463: can just allocate in the first place instead of stack allocating & idup

source/dubregistry/api.d
```diff
@@ -176,6 +188,52 @@ override {
...
+	{
+		if (event == "create") {
+			return postUpdate(_name, secret);
+		} else if (event == "ping") {
+			enforceBadRequest(hook.type == Json.Type.object, "hook is not of type json");
+			auto eventsObj = *enforceBadRequest("events" in hook, "no events object sent in hook object");
+			enforceBadRequest(eventsObj.type == Json.Type.array, "Hook events must be of type array");
+			auto events = eventsObj.get!(Json[]);
+
+			foreach (ev; events)
```
WebFreak001, line 216: can simplify the eventsObj to events and just iterate over Json here

source/dubregistry/api.d
```diff
@@ -176,6 +188,52 @@ override {
...
+				if (ev.type == Json.Type.string && ev.get!string == "create")
+					return "valid";
+
+			string expected = m_registry.getPackageSecret(_name);
+			if (expected.length && secret.length && secret == expected)
+				m_registry.addPackageError(_name,
+					"GitHub hook configuration is invalid. Hook is missing 'create' event. (Tags or branches)");
```
WebFreak001, line 223: this test code will never be run if configuration is valid

source/dubregistry/api.d
```diff
@@ -176,6 +188,52 @@ override {
...
+			auto eventsObj = *enforceBadRequest("events" in hook, "no events object sent in hook object");
+			enforceBadRequest(eventsObj.type == Json.Type.array, "Hook events must be of type array");
+			auto events = eventsObj.get!(Json[]);
+
+			foreach (ev; events)
+				if (ev.type == Json.Type.string && ev.get!string == "create")
+					return "valid";
+
+			string expected = m_registry.getPackageSecret(_name);
+			if (expected.length && secret.length && secret == expected)
```
WebFreak001, line 221: use constant-time compare

PR #416, opened 2019-02-19
Use more than one dummy repositories

wilzbach:
For the Heroku preview, let's use more than one package.

BTW the dummy user's credentials are `dummyuser` and `test1234`.

(this is run for every preview deployment)

### DIFF
--- a/source/app.d
+++ b/source/app.d
@@ -62,11 +62,21 @@ void defaultInit(UserManController userMan, DubRegistry registry)
 		registry.getPackageDump().empty && userMan.getUserCount() == 0)
 	{
 		logInfo("'GENERATE_DEFAULT_DATA' is set and an empty database has been detected. Inserting dummy data.");
-		auto userId = userMan.registerUser("dummy@dummy.org", "dummyUser",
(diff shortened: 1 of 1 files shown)


WebFreak001:
can we also add gitlab and bitbucket repos in the deployment? Would be useful to check that nothing regresses there. I have gitlab repo on https://gitlab.com/WebFreak001/bancho-irc

source/app.d
```diff
@@ -62,11 +62,21 @@ void defaultInit(UserManController userMan, DubRegistry registry)
 		registry.getPackageDump().empty && userMan.getUserCount() == 0)
 	{
 		logInfo("'GENERATE_DEFAULT_DATA' is set and an empty database has been detected. Inserting dummy data.");
-		auto userId = userMan.registerUser("dummy@dummy.org", "dummyUser",
+		auto userId = userMan.registerUser("dummy@dummy.org", "dummyuser",
```
wilzbach: The registry automatically lowercases usernames, but not when searching for them on login. This should just reduce confusion.

wilzbach:
Do we have any repository on Bitbucket?

PR #421, opened 2019-04-07
Use DMD 2.085.1 for the Heroku previews

wilzbach:


wilzbach:
Sticked with DMD for faster preview builds.

PR #422, opened 2019-04-07
Expose /infos endpoint

wilzbach:
This was forgotten in the respective PR (https://github.com/dlang/dub-registry/pull/278).

See also: https://github.com/dlang/dub-registry/pull/278/files#diff-208dd7dc9f2b52b6520f8975b543f926R98

### DIFF
--- a/source/dubregistry/api.d
+++ b/source/dubregistry/api.d
@@ -95,6 +95,7 @@ interface IPackages {
 	@path(":name/:version/info")
 	Json getInfo(string _name, string _version, bool minimize = false);
 
+	@path(":name/:version/infos")
 	Json[string] getInfos(string[] packages, bool include_dependencies = false, bool minimize = false);
 }
 


wilzbach:
CC @s-ludwig as https://github.com/dlang/dub/pull/1366 is over a year old. Any chance we could get this simple fix deployed in the near future?
(the master->stable branch-off for the 2.086 happens around the 15.4)

source/dubregistry/api.d
```diff
@@ -95,6 +95,7 @@ interface IPackages {
 	@path(":name/:version/info")
 	Json getInfo(string _name, string _version, bool minimize = false);
 
+	@path(":name/:version/infos")
 	Json[string] getInfos(string[] packages, bool include_dependencies = false, bool minimize = false);
```
s-ludwig: Shouldn't there be `_name` and `_version` parameters in the signature here? Otherwise the previous endpoint ("/infos") seems to be fine.

s-ludwig: (Previous one is working fine: http://code.dlang.org/api/packages/infos?packages=[%22dub%22])

wilzbach:
Sorry. I was trying to revive https://github.com/dlang/dub/pull/1366 in the middle of the night. Never mind then

PR #423, opened 2019-04-08
fix broken link

### LINKED ISSUE #420: Broken link on package page
Clicking on the "Score" link on a package's page:
score

Leads to this broken link: https://dub.pm/commandline#package-scoring

Caused by: https://github.com/dlang/dub-registry/blob/master/source/dubregistry/viewutils.d#L82

Flying-Toast:
fixes #420

### DIFF
--- a/source/dubregistry/viewutils.d
+++ b/source/dubregistry/viewutils.d
@@ -77,9 +77,9 @@ string formatFuzzyDate(SysTime st)
 	else return format("%s years ago", now.year - st.year);
 }
 
-string formatScore(string rootDir, float score)
+string formatScore(float score)
 {
-	return format(`<a href="%sdevelop#package-scoring">%.1f</a>`, rootDir, score);
+	return format(`<a href="https://dub.pm/develop#package-scoring">%.1f</a>`, score);
 }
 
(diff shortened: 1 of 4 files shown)


source/dubregistry/viewutils.d
```diff
@@ -77,9 +77,9 @@ string formatFuzzyDate(SysTime st)
 	else return format("%s years ago", now.year - st.year);
 }
 
-string formatScore(string rootDir, float score)
+string formatScore(float score)
 {
-	return format(`<a href="%sdevelop#package-scoring">%.1f</a>`, rootDir, score);
+	return format(`<a href="https://dub.pm/develop#package-scoring">%.1f</a>`, score);
```
ZombineDev: This is change is wrong. A website content should always be relative and never hard code the domain name.

s-ludwig: The documentation is now on dub.pm, whereas the main registry is at code.dlang.org, so this is actually a cross-domain link.

ZombineDev: Ah yes, thanks for clarifying that. What I meant is that hard coding the docs domain looks wrong. I think the correct way approach is to read the domain name from a configuration.

Issue #436, opened 2019-08-10
Gitlab dependencies result in server crash


wilzbach:
https://code.dlang.org/packages/silly/0.8.2.zip -> directly leads to a server crash (502)

See also: https://code.dlang.org/packages/silly

AntonMeep:
Can confirm. There aren't any logs or error messages, it just dies for a moment and then gets back up.
Could be an issue with GitLab credentials.

Related issue for silly: https://gitlab.com/AntonMeep/silly/issues/28

AntonMeep:
Full `dub fetch` log:
```
$ dub fetch silly --version=0.8.2 -v
Using dub registry url 'https://code.dlang.org/'
Refreshing local packages (refresh existing: true)...
Looking for local package map at /var/lib/dub/packages/local-packages.json
Looking for local package map at /home/antonmeep/.dub/packages/local-packages.json
No valid package found in current working directory: No package file found in /home/antonmeep/, expected one of dub.json/dub.sdl/package.json
Fetching silly 0.8.2...
Package supplier registry at https://code.dlang.org/ failed with 'Failed to download package silly from https://code.dlang.org/packages/silly/0.8.2.zip', trying fallbacks.
Fallback package supplier registry at https://code-mirror.dlang.io/ failed with 'Failed to download https://code-mirror.dlang.io/api/packages/infos?packages=%5B%22silly%22%5D&include_dependencies=true&minimize=true'.
Fallback package supplier registry at https://dub-registry.herokuapp.com/ failed with 'HTTP request returned status code 503 (Service Unavailable)'.
Failed to download package silly from https://code.dlang.org/packages/silly/0.8.2.zip
```

aliak00:
Has anyone checked out what's the cause of the crash for gitlab deps?

ErnyTech:
I think this unpleasant issue is caused by the change of APIs to download the archive from GitLab (https://gitlab.com/gitlab-com/support-forum/issues/4154)

Now you need to use this: https://docs.gitlab.com/ee/api/repositories.html#get-file-archive

ErnyTech:
I sent a PR: https://github.com/dlang/dub-registry/pull/443

ErnyTech:
The PR has been combined in master and the version of dub-registry hosted on herokuapp is working properly, just try https://dub-registry-staging.herokuapp.com/packages/gitlab-test-package-erny/~master.zip !

[...]

PR #443, opened 2019-10-19
Fix issue #436 - Update GitLab APIs

ErnyTech:
GitLab has updated its API to download the repository archive, so it causes the issue because it could not download the requested repository

Updating the generation of the URL (https://docs.gitlab.com/ee/api/repositories.html#get-file-archive) with the new API is enough to solve the problem

To test: https://salty-hamlet-35363.herokuapp.com/packages/gitlab-test-package/1.0.0.zip

### DIFF
--- a/source/dubregistry/repositories/gitlab.d
+++ b/source/dubregistry/repositories/gitlab.d
@@ -148,7 +148,8 @@ class GitLabRepository : Repository {
 		if (ver.startsWith("~")) ver = ver[1 .. $];
 		else ver = ver;
 		auto venc = () @trusted { return encodeComponent(ver); } ();
-		return m_baseURL.toString()~m_owner~"/"~m_projectPath~"/repository/archive.zip?ref="~venc;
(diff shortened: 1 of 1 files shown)


source/dubregistry/repositories/gitlab.d
```diff
@@ -148,7 +148,7 @@ class GitLabRepository : Repository {
 		if (ver.startsWith("~")) ver = ver[1 .. $];
 		else ver = ver;
 		auto venc = () @trusted { return encodeComponent(ver); } ();
-		return m_baseURL.toString()~m_owner~"/"~m_projectPath~"/repository/archive.zip?ref="~venc;
```
WebFreak001: imo the query parameter "sha" feels like this is passing a wrong argument. Can you add a comment before this line saying that "sha" can be any tag or branch reference too as per GitLab docs? Otherwise it might be questioned in the future

ErnyTech: Ok, done

WebFreak001:
actually the `getAPIURLPrefix()` function should probably be used here?

ErnyTech:
Oh true, I'm sorry I didn't know about the existence of that function

PR #446, opened 2019-11-21
Additional categories

UIMSolutions:
Categories for Cloud and Business apps

### DIFF
--- a/categories.json
+++ b/categories.json
@@ -27,6 +27,16 @@
 				{"name": "game", "description": "Game servers"},
 				{"name": "web", "description": "Web servers"}
 			]},
+			{"name": "business", "description": "Business applications", "categories": [
+				{"name": "sap", "description": "SAP solutions"},
+				{"name": "salesforce", "description": "Salesforce solutions"},
+				{"name": "workday", "description": "Workday solutions"}
+			]},
+			{"name": "cloud", "description": "Cloud solutions", "categories": [
+				{"name": "azure", "description": "Azure cloud solutions"},
+				{"name": "aws", "description": "AWS Azure cloud solutions"},
+				{"name": "google", "description": "Google cloud solutions"}
+			]},
 			{"name": "web", "description": "Web applications/sites", "categories": [
 				{"name": "development", "description": "Development tools"},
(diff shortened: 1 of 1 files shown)


categories.json
```diff
@@ -27,6 +27,16 @@
...
 				{"name": "web", "description": "Web servers"}
 			]},
+			{"name": "business", "description": "Business applications", "categories": [
+				{"name": "sap", "description": "SAP solutions"},
+				{"name": "salesforce", "description": "Salesforce solutions"},
+				{"name": "workday", "description": "Workday solutions"}
+			]},
+			{"name": "cloud", "description": "Cloud solutions", "categories": [
+				{"name": "azure", "description": "Azure cloud solutions"},
+				{"name": "aws", "description": "AWS Azure cloud solutions"},
```
WebFreak001: you put a wrong description here (added Azure)

```suggestion
				{"name": "aws", "description": "Amazon Web Services cloud solutions"},
```

categories.json
```diff
@@ -27,6 +27,16 @@
 				{"name": "game", "description": "Game servers"},
 				{"name": "web", "description": "Web servers"}
 			]},
+			{"name": "business", "description": "Business applications", "categories": [
+				{"name": "sap", "description": "SAP solutions"},
+				{"name": "salesforce", "description": "Salesforce solutions"},
+				{"name": "workday", "description": "Workday solutions"}
```
WebFreak001: how about putting all of these 3 categories in a parent category called `"ERP software"` or similar

WebFreak001 (changes requested):
I don't think it's a good idea to include brand names in the categories list.

This also currently fails to adhere to the trademark usage guidelines of some products, as it also needs notes about for example `"<SAP TRADEMARKS> is/are the trademark(s) or registered trademark(s) of SAP SE in Germany and in several other countries."` and the &reg; or &trade; symbols depending on the name.

I'm for adding `cloud`, `business` and probably also `erp` but I don't think it's a good idea to include brand names. Also there aren't that many libraries and applications fitting in these categories currently so I think the broad categories are enough to be descriptive.

PR #447, opened 2019-11-23
Work around a stack corruption causing broken GitLab downloads

### LINKED ISSUE #349: Internal server error for projects hosted via gitlab

### LINKED ISSUE #436: Gitlab dependencies result in server crash

s-ludwig:
Returning the `Nullable!DbPackage`, implicitly converted to `DbPackage`, triggered a codegen issue on LDC, leading to the version string to become empty on Windows and causing a crash on the following method call on Linux.

Closes #436. Fixes #349.

### DIFF
--- a/source/dubregistry/dbcontroller.d
+++ b/source/dubregistry/dbcontroller.d
@@ -136,7 +136,7 @@ class DbController {
 	{
 		auto pack = m_packages.findOne!DbPackage(["name": packname]);
 		enforce!RecordNotFound(!pack.isNull(), "Unknown package name.");
-		return pack;
+		return pack.get;
 	}
 
 	auto getPackages(scope string[] packnames...)
@@ -156,7 +156,7 @@ class DbController {
 	{
 		auto pack = m_packages.findOne!DbPackage(["_id": id]);
 		enforce!RecordNotFound(!pack.isNull(), "Unknown package ID.");
-		return pack;
+		return pack.get;
 	}
 
 	auto getAllPackages()


source/dubregistry/registry.d
```diff
@@ -388,7 +388,9 @@ class DubRegistry {
 
 	void downloadPackageZip(string packname, string vers, void delegate(scope InputStream) @safe del)
 	{
+		assert(vers.length > 0);
 		DbPackage pack = m_db.getPackage(packname);
+		assert(vers.length > 0);
```
WebFreak001: is it intended that this is 2x the same assert with no changes in value? There is nothing accessing vers there between the 2 calls.

If the stack got corrupted here it could possibly have any value other than 0 too so the assert would probably rarely trigger here. (the `length > 0` is the same as `length != 0` because it's unsigned so it's just checking for only 0) Also it's confusing for future code review why this assert exists 2x

s-ludwig: I agree on it being kind of confusing, maybe I'll just remove it. Of course *anything* can happen, like on Linux, where the symptoms were completely different. Aith the `alias this` for `Nullable.get` being removed soon, this is unlikely to slip back in, so the assertion doesn't have much chance of being an actual help in the future.

s-ludwig:
I've also deployed these changes to code.dlang.org and the gitlab test package now downloads without errors.

aliak00:
Omg why doesn't Nullable just die. Someone kill it with fire!

Issue #393, opened 2018-11-28
Cannot register/update package from a private GitLab repository


the-Mitu:
The error I get:

    Failed to find a branch containing a valid package description file:
    master: dub.json(1): Error: Expected valid JSON token, got '<!DOCTYPE ht'. 

I **guess** GitLab responds with login page. GitLab URL and access token are set up in the configuration file.

Registering public project works properly.

Geod24:
Fixed by #452

Issue #117, opened 2015-07-31
Add support for more git hosts


SuperDoxin:
currently only github and bitbucket are supported. I would love to see at least gitlab.com supported too, but ideally any git url ought to work.

profan:
Supporting this as well, seems like it would be natural to allow any given git remote.

s-ludwig:
The problem with arbitrary git URLs is mainly the download link. It would require downloading, extracting, zipping and storing all of such packages locally on the registry server. Currently this wouldn't be an issue, but if we ever grow to the numbers of packages that the most popular package systems have today, it will require an upgrade of the hardware infrastructure. Another potential issue is that every git repository requires its own server connection, while the HTTP hoster protocols can reuse the same connection for many consecutive requests. So this would create a lot of overhead for both, the registry and the hoster.

So, this is certainly doable, but doesn't scale well enough for the current server infrastructure.

GitLab support is basically there, but so far I've only tested it with a local GitLab instance. I'll switch that live probably in about one or two weeks.

SuperDoxin:
so why isn't the git protocol being used here? seems to me that the dub client could just git clone whatever it needs instead of the server having to hand it a zipfile.

s-ludwig:
Using the git protocol would mean either that the server would have to mirror all repositories, while it now only has to store the meta data, or the client would also have to use either libgit or the git command line client, both of which is problematic in one way or another. This would also mean that git gets special treatment - but we currently also support HG repositories on Bitbucket for example, because everything works using HTTP.

sigod:
I too would love to see GitLab support.

s-ludwig:
I've added a branch where the code I once wrote for GitLab is integrated: https://github.com/D-Programming-Language/dub-registry/tree/gitlab-support

Would be good if someone could test this and provide patches if necessary. I'm currently lacking the resources to finish this and we have internally switched to Gogs now, so when I get the time I will concentrate on that instead.

sigod:
> I'm currently lacking the resources to finish this and we have internally switched to Gogs now, so when I get the time I will concentrate on that instead.

Very interesting service. Are there any major hosting for it? I don't see how it can be supported for user hosted packages.

s-ludwig:
Gogs would indeed just be for private/internal hosting, I don't know of any public service, although I haven't searched for it either. I want to set up a local dub-registry, so that building projects just works, without having to manually clone all dependencies.

BlackEdder:
[...]

PR #452, opened 2020-01-15
Fixed GitLab private repositories download file URL

vitalka200:
Currently GitLab support is broken for private repositories.
see: https://gitlab.com/gitlab-org/gitlab-foss/issues/55081

This commit makes dub-registry use GitLab api thus fixing private repository browsing


vitalka200:
Yep, for both public/private repos. Looks like this one #393 can be closed and this one #117 at least updated that registry now supports GitLab

Geod24:
Thanks!

PR #468, opened 2020-06-24
Custom mail settings and related changes

s-ludwig:
Also upgrades to the latest vibe.d and UserMan versions to fix compiling with current frontend versions and fixes certificate related issues.

This change is required to make password reset e-mails work again, since the server running the registry has no local mail server anymore.

### DIFF
--- a/dub.sdl
+++ b/dub.sdl
@@ -6,9 +6,9 @@ authors "Sönke Ludwig" "Martin Nowak" "Anton Fediushin aka ohdatboi" \
 	"see GitHub for more"
 license "BSL-1.0"
 
-dependency "vibe-d" version="~>0.8.6"
+dependency "vibe-d" version="~>0.9.0-beta.1"
 dependency "dub" version="~>1.21.0"
-dependency "userman" version="~>0.3.8"
+dependency "userman" version="~>0.4.0"
 subConfiguration "dub" "library-nonet"
 
 versions "VibeJsonFieldNames"
--- a/source/app.d
+++ b/source/app.d
@@ -18,6 +18,7 @@ import std.algorithm : sort;
 import std.process : environment;
 import std.file;
 import std.path;
+import std.traits : isIntegral;
 import userman.db.controller : UserManController, createUserManController;
 import userman.userman : UserManSettings;
 import userman.web;
@@ -87,10 +88,20 @@ struct AppConfig
 	string ghuser, ghpassword,
 		   glurl, glauth,
 		   bbuser, bbpassword;
+
+	string mailServer;
+	ushort mailServerPort;
+	SMTPConnectionType mailConnectionType;
+	string mailClientName;
+	string mailUser;
+	string mailPassword;
+
+
 	void init()
 	{
 		import dub.internal.utils : jsonFromFile;
 		auto regsettingsjson = jsonFromFile(NativePath("settings.json"), true);
+		// TODO: use UDAs instead
 		static immutable variables = [
 			["ghuser", "github-user"],
 			["ghpassword", "github-password"],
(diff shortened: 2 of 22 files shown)


source/app.d
```diff
@@ -168,9 +202,28 @@ void main()
 		auto udbsettings = new UserManSettings;
 		udbsettings.serviceName = "DUB - The D package registry";
 		udbsettings.serviceURL = URL("http://code.dlang.org/");
-		udbsettings.serviceEmail = "noreply@vibed.org";
+		udbsettings.serviceEmail = "noreply@rejectedsoftware.com";
```
WebFreak001, line 205: wouldn't this be better configurable or `@code.dlang.org`?

source/app.d
```diff
@@ -85,25 +88,54 @@ struct AppConfig
...
+	string mailClientName;
+	string mailUser;
+	string mailPassword;
+
+
 	void init()
 	{
 		import dub.internal.utils : jsonFromFile;
 		auto regsettingsjson = jsonFromFile(NativePath("settings.json"), true);
+		// TODO: use UDAs instead
```
WebFreak001, line 106: JSON with environment fallback might also be a common use-case we would want to support in vibe.d

source/app.d
s-ludwig: Configurable definitely makes sense. But `@code.dlang.org` doesn't have a mail server configured, so I just resorted to something that was already there.

s-ludwig:
@wilzbach: Unfortunately I can't see what went wrong with the heroku deployment, as it requires to log in, can you forward the error?

dub.sdl
```diff
@@ -6,9 +6,9 @@ authors "Sönke Ludwig" "Martin Nowak" "Anton Fediushin aka ohdatboi" \
 	"see GitHub for more"
 license "BSL-1.0"
 
-dependency "vibe-d" version="~>0.8.6"
+dependency "vibe-d" version="~>0.9.0-beta.1"
```
Geod24, line 9: That's why we really need a new release for Vibe :)

source/app.d
```diff
@@ -114,7 +114,7 @@ void main()
 	bool noMonitoring;
 	setLogFile("log.txt", LogLevel.diagnostic);
 
-    version (linux)
+    version (linux) version (DMD)
```
Geod24: Personally I went with a slightly different approach: https://github.com/dlang/dlang-bot/commit/9b0e9cb6b142b8c660c659bfa32bc31e5ae23365

s-ludwig: For me this resulted in a linker error rather than a compile error, so the `typeof` wouldn't have helped. Maybe this has changed in a recent LDC version?

s-ludwig:
> Not so savvy about the code but everything looks good from here.

Agreed, I didn't have an eye on the recent changes, but it looks like a lot of not-quite-clean code has accumulated. For this PR though, I just tried to fit in the new functionality without investing additional time for refactoring.

I'd merge this now, despite the Heroku issue, since that shouldn't be a blocker for anything AFAICS.

s-ludwig:
Okay, the new version is now deployed and password reset mails should work again.

wilzbach:
> @wilzbach: Unfortunately I can't see what went wrong with the heroku deployment, as it requires to log in
> I do not have access to Heroku either, so can't help with the deployment issue.

@Geod24: added you to the dub-registry-staging repo. @WebFreak001 and @s-ludwig I couldn't find a Heroku account for you (or maybe you don't have one?), so if you want to have access, just shoot me an email with your Heroku account.

Anyhow, it looks like DMD is running out of memory:

```
       source/dubregistry/web.d(471,19): Deprecation: constructor `userman.web.UserManWebAuthenticator.this` is deprecated
       Compiling Diet HTML template search_results.dt...
       Compiling Diet HTML template my_packages.dt...
       Compiling Diet HTML template my_packages.register.dt...
       Compiling Diet HTML template my_packages.register.warn_fork.dt...
       Compiling Diet HTML template my_packages.package.dt...
       Compiling Diet HTML template my_packages.remove.dt...
       dmd failed with exit code 1.
 !     Push rejected, failed to compile D app.
```

There were a few improvements recently, so I made a PR to bump the DMD version (https://github.com/dlang/dub-registry/pull/469).

PR #469, opened 2020-06-25
Bump dmd to 2.092.1

wilzbach:


Geod24:
Can't we compile with LDC ?

wilzbach:
> Can't we compile with LDC ?

Considering that this is still failing on Heroku, we have to.

Geod24:
I don't see it in my apps BTW

WebFreak001:
if the issue is something like
vibe.d build issues
then you can fix it by running
`rm **/dub.selections.json` before building

Issue #442, opened 2019-10-13
Bring back good-old copy&paste dependency boxes


0xEAB:
The way #391 forces you to use to add new dependencies blows up all formatting and structuring of your `dub.[json|sdl]`. This is definitely not acceptable for quite a few projects/use-cases.

I wonder why such changes always have to be applied in a black/white-only way. Why can't there be **both** options? What would be the problem with having all of them: the fresh *`dub add` box* and the good old *copy&paste boxes*?

(This issue is *not* a duplicate of #439 because #439 aims at improving `dub add` and/or the *`dub add` box* instead.)

0xEAB:
UPDATE: This is actually **pretty annoying**. You might not believe it, but distros out there package older versions of DUB where `dub add` isn't a thing yet. (There you'll find yourself doing all this busywork of trying to figure out which version to specify plus how that syntax was.)

PetarKirov:
> Why can't there be both options? What would be the problem with having all of them: the fresh dub add box and the good old copy&paste boxes?

I agree that having both options is a good idea. Additionally we need to fix `dub add` rewrite `dub.{sdl,json}` files from scratch, but instead just modifies the dependencies section in a smarter way.


> You might not believe it, but distros out there package older versions of DUB where dub add isn't a thing yet.

`<rant>`
Supporting the D toolchain (or any other language for that matter) in a non-rolling distro is a non-starter in my opinion. Due to the release cadence of distros like Debian or Ubuntu, all language toolchains are always outdated. This may not be an issue with C, where new versions are released not less than 10 years apart, so it's unlikely that you're going to miss any feature, but D the toolchain has a major release every two months.
To add insult to the injury, typically distro package managers have their own way of building packages (not using the upstream release scripts), so it's hard to know exactly what is being packaged.
Say we need to rearchitect the API of the dub registry. When users use rolling distros, or even better - manage their toolchains themselves, we can ask everyone to upgrade and wait a couple of releases, before enabling the new API endpoints and shutting down the old ones. But now with non-rolling distros in the picture, who knows how many years we would need if we want to make breaking changes to the API registry?
`</rant>`

Geod24:
> who knows how many years we would need if we want to make breaking changes to the API registry?

My turn to `<rant>`.
[...]

PR #474, opened 2020-07-24
Bring back copy&paste dependency boxes

0xEAB:
Implements #442.

Thanks to @WebFreak001 for helping me getting a local copy running (and making this PR possible).

### DIFF
--- a/views/view_package.dt
+++ b/views/view_package.dt
@@ -72,6 +72,22 @@ block body
 				input(id="package-clipboard",value='dub add #{packageName}')
 				button(class="btn-clipboard",data-clipboard-target="#package-clipboard")
 					img(src="/images/clippy.svg", width=13, alt="Copy to clipboard")
+		.clipboard-manual
+			p
+				strong Manual usage
+				br
+				| Put the following dependency into your project's dependences section:
+			.clipboard-pkg
+				.clipboard-pkg-row
+					label.clipboard-pkg-title(for="package-clipboard-json") dub.json
+					input(id="package-clipboard-json",value='"#{packageName}": "#{expr}"')
+					button(class="btn-clipboard",data-clipboard-target="#package-clipboard-json")
+						img(src="/images/clippy.svg", width=13, alt="Copy to clipboard")
+				.clipboard-pkg-row
+					label.clipboard-pkg-title(for="package-clipboard-sdl") dub.sdl
+					input(id="package-clipboard-sdl",value='dependency "#{packageName}" version="#{expr}"')
+					button(class="btn-clipboard",data-clipboard-target="#package-clipboard-sdl")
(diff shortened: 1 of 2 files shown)


views/view_package.dt
```diff
@@ -72,6 +72,23 @@ block body
 				input(id="package-clipboard",value='dub add #{packageName}')
 				button(class="btn-clipboard",data-clipboard-target="#package-clipboard")
 					img(src="/images/clippy.svg", width=13, alt="Copy to clipboard")
+		div
```
WebFreak001, line 75: empty useless diff

0xEAB:
## Screenshot

Screenshot from 2020-07-24 20-51-42

views/view_package.dt
```diff
@@ -72,6 +72,23 @@ block body
...
+		.clipboard-manual
+			p
+				strong Manual usage
+				br
+				| Put the following dependency into your project's dependences section:
+			.clipboard-pkg
+				.clipboard-pkg-row
+					.clipboard-pkg-title dub.json
```
WebFreak001, line 83: this is a label, make it a `label` tag and add `(for="package-clipboard-json")` (and adjust CSS to support label instead of div here)

When you click a label the user by default even selects the input, which can be convenient for the user and is more accessible instead of just being some text before a text box.

views/view_package.dt
```diff
@@ -72,6 +72,23 @@ block body
...
+					input(id="package-clipboard-json",value='"#{packageName}": "#{expr}"')
+					button(class="btn-clipboard",data-clipboard-target="#package-clipboard-json")
+						img(src="/images/clippy.svg", width=13, alt="Copy to clipboard")
+				.clipboard-pkg-row
+					.clipboard-pkg-title dub.sdl
```
WebFreak001, line 88: ditto

views/view_package.dt
0xEAB: well, it groups this thingy, and it's what I applied the margins (top+bottom) to

WebFreak001: no you are missing an indentation level if you want this to group anything

0xEAB: ah, I see…
just writing a class without any element type doesn't make it apply to the "current" one but automagically creates a new `div`

WebFreak001:
if there are no comments at all in the next 2 days I will add the `72h no objections -> merge` label because I would like to see this get in (and make a follow up PR with design changes which can get more discussion time than the feature itself)

Geod24:
I don't see a reason to delay the merge, dub-registry doesn't see enough contributors to have that luxury, and the improvement is quite non-controversial.

wilzbach:
> I don't see a reason to delay the merge

FYI: the dub-registry isn't auto-deployed. Only Sönke can deploy it manually, so this might take a few days/weeks to appear live.

PR #478, opened 2020-08-28
support up to 1000 tags per package

9il:
Required to support mir-algorithm in private codebases


wilzbach:
Unfortunately I don't think the fix is that easy as AFAICT GitHub limits `per_page` to a maximum of 100.
To receive more tags, pagination is required.

https://docs.github.com/en/rest/guides/traversing-with-pagination

9il:
new MR in #479

PR #479, opened 2020-08-29
fetch all github tags

9il:


source/dubregistry/repositories/github.d
```diff
@@ -44,28 +44,35 @@ class GithubRepository : Repository {
...
+					auto tagname = tag["name"].get!string;
+					Json commit = readJson(getAPIURLPrefix()~"/repos/"~m_owner~"/"~m_project~"/commits/"~tag["commit"]["sha"].get!string, true, true);
+					ret ~= RefInfo(tagname, tag["commit"]["sha"].get!string, SysTime.fromISOExtString(commit["commit"]["committer"]["date"].get!string));
+					logDebug("Found tag for %s/%s: %s", m_owner, m_project, tagname);
+				} catch( Exception e ){
+					throw new Exception("Failed to process tag "~tag["name"].get!string~": "~e.msg);
+				}
 			}
+			if (count < 100)
+				break;
```
wilzbach, line 68: Unfortunately that's not the right way to do pagination as you might send an unnecessary request if the count is 100.
You need to look at the "link" header.

```
$ curl -v "https://api.github.com/repos/dlang/dmd/tags?per_page=1" 
...
< link: <https://api.github.com/repositories/1257070/tags?per_page=1&page=2>; rel="next", <https://api.github.com/repositories/1257070/tags?per_page=1&page=258>; rel="last"
```

source/dubregistry/repositories/github.d
```diff
@@ -44,28 +44,35 @@ class GithubRepository : Repository {
...
-				auto tagname = tag["name"].get!string;
-				Json commit = readJson(getAPIURLPrefix()~"/repos/"~m_owner~"/"~m_project~"/commits/"~tag["commit"]["sha"].get!string, true, true);
-				ret ~= RefInfo(tagname, tag["commit"]["sha"].get!string, SysTime.fromISOExtString(commit["commit"]["committer"]["date"].get!string));
-				logDebug("Found tag for %s/%s: %s", m_owner, m_project, tagname);
-			} catch( Exception e ){
-				throw new Exception("Failed to process tag "~tag["name"].get!string~": "~e.msg);
+		for (size_t page = 1; ; page++)
+		{
+			Json tags;
+			try tags = readJson(getAPIURLPrefix()~"/repos/"~m_owner~"/"~m_project~"/tags?per_page=100&page=" ~ page.text);
```
wilzbach, line 53: I wonder why `git ls-remote` hasn't been used. It has the big advantage that it is not rate-limited (and even repository provider independent)

```
git ls-remote --tags https://github.com/libmir/mir "refs/tags/v*"
```

wilzbach (commented):
I guess the biggest concern I have is that this might exhaust the registries rate-limit even faster, but I guess the problem is that for every tag there's a call to the GitHub API in order to retrieve the commit date.

source/dubregistry/repositories/github.d
9il: This will affect less the %1 of repositories and with a probability less than %1. In total it affects less with the probability less than %0.01.

wilzbach: Yes, but it will still result in an error at the dub registry.

WebFreak001:
same as #458 ?

source/dubregistry/repositories/github.d
9il: Why error?
https://api.github.com/repos/libmir/mir-algorithm/tags?per_page=100&page=10 returns an empty array and status 200.

9il:
> same as #458 ?

similar. This one is straightforward, unoptimized, and slim.

source/dubregistry/repositories/github.d
9il: maybe, I just need the registry work for mir ASAP and without new surprises. #458 looks better for optimization purpose.

wilzbach (approved):

wilzbach:
Fair enough, I don't want to block progress. Though note that this isn't auto-deployed and I don't have access to the registry server. You would need to ping Sönke about it.

PR #481, opened 2020-10-12
Enable partial search

Imperatorn:
**Problem**
When searching for packages, partial maches are not found.

**Example**
Search for "jwt" on dub.pm -> https://code.dlang.org/search?q=jwt

**Result**
Does not find jwtd, only jwt

**Proposed solution**
Use partial search by changing query

**Possible alternative solution**
Configure partial search

**Possible alternative solution 2**
Enable usage of wildcard operator

**NOTE**
Untested solution

### DIFF
--- a/source/dubregistry/dbcontroller.d
+++ b/source/dubregistry/dbcontroller.d
@@ -331,7 +331,7 @@ class DbController {
 		}
 
 		auto pkgs = m_packages
-			.find(["$text": ["$search": query]], ["textScore": bson(["$meta": "textScore"])])
+			.find(["$text": ["$search": "/" ~ query ~ "/i"]], ["textScore": bson(["$meta": "textScore"])])
 			.sort(["textScore": bson(["$meta": "textScore"])]) // sort to only keep most relevant results
 			.limit(50) // limit irrelevant sort results (fixes #341)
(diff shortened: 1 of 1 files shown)


Imperatorn:
Is the docker build broken?

error building image: error building stage: failed to get files used from context: failed to get fileinfo for /workspace/dub-registry: lstat /workspace/dub-registry: no such file or directory

Imperatorn:
Hmm, proposed solution might not work due to mongodb (never used it myself) does not do searches that way.
Maybe someone can use any of the other solutions or implement a partial search?

Imperatorn:
Is partial search possible?

s-ludwig:
AFAIK, MongoDB doesn't explicitly support partial searches. But what we could do is to manually perform a text search within the sorted list of package names and prioritize that up to a certain degree over the MongoDB results. That way we could implement any kind of partial matching and this would also solve the issue that text search prioritization is currently broken (the package name should have higher priority than e.g. the README contents, but currently this is typically not the case).

dejlek:
This is implemented by https://github.com/dlang/dub-registry/pull/600 , and tested.

WebFreak001:
closing in favor of #600

PR #486, opened 2020-12-21
Mirror improvements

s-ludwig:
This mainly allows to run a mirror in a way where the background service only provides the web frontend (and REST API) and database updates happen in separate process invocations (executed by a cron job). Since there is currently a massive memory leak somewhere in the "request package dump -> read compressed answer and uncompress -> deserialize JSON to native representation" process.

This is already deployed on codemirror.dlang.org, so that it's uptime should now also be virtually 100%.

### DIFF
--- a/source/app.d
+++ b/source/app.d
@@ -147,7 +147,7 @@ struct AppConfig
 
 void main()
 {
-	bool noMonitoring;
+	bool noMonitoring, noServe;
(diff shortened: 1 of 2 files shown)


wilzbach (approved):

wilzbach:
Not sure, what this Docker build is about, but the failure looks unrelated.

PR #493, opened 2021-04-02
Add BetterC category

dkorpel:
There are currently 38 packages mentioning `-betterC` support in the description, and an unknown number of packages supporting `-betterC` not mentioned in the description. Adding a category for it could make it easier to find them once package maintainers apply it appropriately.

### DIFF
--- a/categories.json
+++ b/categories.json
@@ -36,6 +36,7 @@
 		{"name": "library", "description": "Development library", "categories": [
 			{"name": "std_aspirant", "description": "Candidate for inclusion in the D standard library"},
 			{"name": "audio", "description": "Audio libraries"},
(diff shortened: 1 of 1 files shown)


WebFreak001:
Hm I can see you put it sorted into the list, but there might also be benefit in putting it next to similar categories as currently not the whole list is sorted. What do others think how should this be sorted?

1. Alphabetically like it (partially) is now?
2. Next to similar categories like `@nogc`

react :tada: for 1, react :rocket: for 2

dkorpel:
I didn't consider that the order in the json file would affect the presentation order. In that case, I would put it close to related categories.

WebFreak001 (approved):
comment reactions indicate this sorting is ok

Issue #495, opened 2021-05-06
Manual update does not work


matti-82:
I triggered a manual update for the package easyd, but it is never performed. I see the status message "Still have to wait for 1683 other packages to complete". And when the countdown goes down to ~1000, it jumps up to ~1700 again.

s-ludwig:
Could it be that it has actually performed the update, but was too quick to notice? I just tried and it definitely worked, but it was quite quick due to the fact that the package that was already updating at that moment only had few versions.

It might make sense to add a "last updated" date to each package, so that this becomes clearer.

matti-82:
0.9.0 is still shown, although 0.13.0 is the current version on github

matti-82:
Found the problem, the tags have not been pushed.

PR #496, opened 2021-05-07
Ensure fair package updates and fix queue overflow

s-ludwig:
There were several issues playing together here, resulting in some packages never getting updated:

- The periodic package update was scheduling packages redundantly, combined with an update cycle that is nowadays longer than the anticipated 30 minutes, this resulted in the back part of the package list never being reached
- The secondary consequence was that the process crashed at some point with an assertion due to the update queue filling up, resulting in another possible reason for the package list to never be fully processed
- Packages were always inserted into the queue in the same order - now they are randomized, resulting in a natural robustness against similar issues

See also dlang/dub#2128 and #495.

### DIFF
--- a/source/dubregistry/registry.d
+++ b/source/dubregistry/registry.d
@@ -444,7 +444,8 @@ class DubRegistry {
 		auto allpacks = this.availablePackages.array;
 		randomShuffle(allpacks);
 		foreach (packname; allpacks)
-			triggerPackageUpdate(packname);
+			if (!m_updateQueue.isPending(packname))
+				triggerPackageUpdate(packname);
 	}
 
 	protected string validateRepository(DbRepository repository)
--- a/source/dubregistry/registry.d
+++ b/source/dubregistry/registry.d
@@ -431,10 +431,19 @@ class DubRegistry {
 
 	void updatePackages()
 	{
+		import std.random : randomShuffle;
+
 		logDiagnostic("Triggering package update...");
 		// update stat distributions before score packages
 		m_statDistributions = m_db.getStatDistributions();
-		foreach (packname; this.availablePackages)
+
+		// shuffle the package list to ensure a uniform probability of each
+		// package getting chosen over time, regardless of how long it takes
+		// to process the whole list and whether the process gets restarted
+		// in-between
+		auto allpacks = this.availablePackages.array;
+		randomShuffle(allpacks);
+		foreach (packname; allpacks)
 			triggerPackageUpdate(packname);
 	}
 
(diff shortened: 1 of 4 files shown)


source/dubregistry/registry.d
```diff
@@ -431,11 +431,21 @@ class DubRegistry {
...
-		foreach (packname; this.availablePackages)
-			triggerPackageUpdate(packname);
+
+		// shuffle the package list to ensure a uniform probability of each
+		// package getting chosen over time, regardless of how long it takes
+		// to process the whole list and whether the process gets restarted
+		// in-between
+		auto allpacks = this.availablePackages.array;
+		randomShuffle(allpacks);
+		foreach (packname; allpacks)
```
WebFreak001, line 446: instead of allocating, you could use the randomCover function from std.random

source/dubregistry/registry.d
```diff
@@ -431,11 +431,21 @@ class DubRegistry {
...
+			if (!m_updateQueue.isPending(packname))
```
WebFreak001, line 447: shouldn't this be handled by triggerPackageUpdate?

s-ludwig: `triggerPackageUpdate` is also used for the immediate update trigger, so that would then require further changes.

source/dubregistry/registry.d
s-ludwig: Turns out it still has to allocate, because `randomCover` requires a random access range.

WebFreak001: ah yeah makes sense, probably doesn't make sense to implement random access range on the cursor object because it would probably worsen performance.

I think `randomCover` also allocates a list of indices to iterate over so randomShuffle is probably the best option to use here. Maybe include a comment about this to avoid future code review thinking randomCover could be used instead?

Imperatorn:
Merge?

skoppe:
Would love to have this. Currently have to wait days to get tags in for the [concurrency](https://code.dlang.org/packages/concurrency) package. Needless to say it blocks us.

WebFreak001 (approved):

mingwugmail:
Can we rollout this fix to production server asap?

s-ludwig:
Deployed now. I'll check the log in a few hours to see whether packages are updated consistently in each cycle. BTW, I'll also increase the update cycle from 30 minutes to one hour, since that is not the reality anymore. Using OAuth and update hooks is really getting essential now to get timely updates.

s-ludwig:
Looks like all packages are updated consistently now, averaging about one hour between updates.

CyberShadow:
@s-ludwig Could you please have a look at https://code.dlang.org/packages/emsi_containers , it hasn't been updated in over a year.

s-ludwig:
@CyberShadow Looking into it. I've added some additional logging for now, because there were two code paths where a package could get (almost) silently skipped on error, and will observe what happens during the next update cycles.

PR #501, opened 2021-07-03
Add OpenSearch description

kubo39:
This allows to search dub packages easier from address bar in browsers.

### DIFF
--- a/public/opensearch.xml
+++ b/public/opensearch.xml
@@ -3,5 +3,5 @@
   <ShortName>DUB</ShortName>
   <Description>Search for dub packages</Description>
   <Image type="image/png">https://code.dlang.org/images/logo-small.png</Image>
-  <Url type="text/html" method="get" template="https://code.dlang.org/search?q={search-query}"/>
+  <Url type="text/html" method="get" template="https://code.dlang.org/search?q={searchTerms}"/>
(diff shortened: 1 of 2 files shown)


public/opensearch.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/">
+  <ShortName>DUB</ShortName>
+  <Description>Search for dub packages</Description>
+  <Image type="image/png">https://code.dlang.org/images/logo-small.png</Image>
+  <Url type="text/html" method="get" template="https://code.dlang.org/search?q={search-query}"/>
```
s-ludwig, line 6: I'm not familiar with the OpenSource format so I may be wrong on this, but [it looks](https://github.com/dewitt/opensearch/blob/master/opensearch-1-1-draft-6.md#the-searchterms-parameter) like this should be "{searchTerms}":

```suggestion
  <Url type="text/html" method="get" template="https://code.dlang.org/search?q={searchTerms}"/>
```

kubo39: Thanks, I missed.

s-ludwig (approved):

PR #504, opened 2021-07-04
make dpldocs links versioned instead of master

WebFreak001:
This fixes that you need to keep reloading the cache to see updated documentation and allows you to view documentation of older versions.


Geod24 (approved):
Low cost high ROI, love it

WebFreak001:
would love if this could be deployed soon :)

PR #515, opened 2022-02-11
Add GitHub Actions

kubo39:


thewilsonator:
This has no tests. Are they only activated once pulled?

WebFreak001:
> This has no tests. Are they only activated once pulled?

That's correct for non-members I think, but you can see the results on their branch: https://github.com/kubo39/dub-registry/tree/github-actions

thewilsonator (approved):

PR #522, opened 2022-05-11
Add support for administrator users.

s-ludwig:
Adds a new field "administrators" to the settings file to configure a list of admin user names. Administrators have access rights for all packages instead of just their own.


Geod24 (approved):
That might unblock a few issues, thanks!

PR #532, opened 2022-10-18
dbcontroller: Remove old migration code

Geod24:
Was looking at adding a migration myself, and obviously there is a bit of legacy that we can (and should) clean up first.

### DIFF
--- a/source/dubregistry/dbcontroller.d
+++ b/source/dubregistry/dbcontroller.d
@@ -38,9 +38,6 @@ class DbController {
 		// migrations:
 		//
 
-		// remove old logo fields
-		m_packages.update(["logoHash": ["$exists": true]], ["$unset": ["logo": 0, "logoHash": 0]], UpdateFlags.multiUpdate);
-
 		// create indices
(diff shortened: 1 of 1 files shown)


source/dubregistry/dbcontroller.d
```diff
@@ -38,54 +38,6 @@ class DbController {
...
-
-		// default initialize missing scores with zero
-		float score = 0;
-		m_packages.update(["stats.score": ["$exists": false]], ["$set": ["stats.score": score]], UpdateFlags.multiUpdate);
-
-		// remove old logo fields
-		m_packages.update(["logoHash": ["$exists": true]], ["$unset": ["logo": 0, "logoHash": 0]], UpdateFlags.multiUpdate);
-
-		// create indices
-		m_packages.ensureIndex([tuple("name", 1)], IndexFlags.Unique);
```
WebFreak001, line 81: you are dropping indices here, I don't think these should be dropped

WebFreak001 (changes requested):

Geod24:
@WebFreak001 : Rebased, fixed

WebFreak001 (approved):

PR #545, opened 2023-06-25
Use Configy to read configuration, let env override config

Geod24:
Normally, the environment variables always take preference on the configuration.
Hence, changing the priority here makes sense. Additionally, Dub already vendors
Configy, which we can just use instead of manual parsing.


thewilsonator (approved):

s-ludwig:
Is the `@Name` something provided by Configy? Or more importantly, does it expect the camelCase names in the JSON file, or the `@Name` ones? It's not really visible from the source code now, so at least a comment, or better yet a test, is probably a good idea here.

Geod24:
> Is the `@Name` something provided by Configy? Or more importantly, does it expect the camelCase names in the JSON file, or the `@Name` ones? It's not really visible from the source code now, so at least a comment, or better yet a test, is probably a good idea here.

It expects the `Name` one. `Name` and `Optional` are in the attributes module.
Moved the code to a new module (since `dub test` ignores unittests in the `app` module) and added two tests covering hopefully all features.

s-ludwig:
Okay, I see I got the previous behavior backwards, so this is good!

PR #550, opened 2023-08-14
implement sharing packages to other users

### LINKED ISSUE #536: Multiple owners / admins for packages
The DUB registry should support adding multiple DUB user accounts as owners to packages, to avoid issues such as #535 and #459

Also interesting would be having group / organization accounts (could even be linked to github organizations or teams)

There should probably also be the possibility to transfer package ownership by users themself (with recipient needing to accept, to not be able to spam users with foreign garbage packages)

WebFreak001:
fixes #536

I didn't add transferring package ownership, since I wanted to rather do that with an "accept" system, so that you won't suddenly become owner of random spam packages, and felt like that was out of scope / too big here.

permissions view
permissions view on mobile
my packages view

only having partial permissions (only trigger manual update):

partial permissions

leave package page:

leave package page


source/dubregistry/web.d
```diff
@@ -305,6 +305,18 @@ class DubRegistryWebFrontend {
 				return url;
 			}
 
+			bool hasManagementAccess;
+			bool isMine;
+			if (req.session) {
+				auto userId = User.ID.fromString(req.session.get!string("userID"));
+				hasManagementAccess = packinfo.hasPermissions(
```
MrcSnm, line 312: I don't know much about the context of those files but this line looks a little strange to me.

MrcSnm: Btw, other than that, I didn't noticed anything strange in the code.

WebFreak001: this is used to show the "Edit this package" button on the view package page - since we potentially don't have a user, we need to fetch it from the session here. The userId / session reading code is just moved from inside the .dt template into the .d file here. The first hasPermissions check checks if we should show the button, the second checks if the package is owned by the logged in user (to show a different text below the button)

MrcSnm: Oh yea, what I found strange was using the `readonly` permission for hasManagementAccess, it looked very strange.

WebFreak001:
ping @s-ludwig

s-ludwig (approved):
I'm pretty sure that I already reviewed this and was basically okay with everything, but got distracted and forgot to finish the review. The only thing that would be nice to use here is the role functionality of vibe.web's authorization system, but we can target that for a later refactoring.

WebFreak001:
thanks! any chance this gets deployed soon?

s-ludwig:
I'll upload a new build! Can you have a quick look at the two additions in #553? I could then also look into whether we can enable gitea.com public repositories.

s-ludwig:
Latest master is deployed now.

WebFreak001:
the CSS changes don't seem to have deployed, possibly indicating that also the JS changes might not be deployed.

Issue #344, opened 2018-06-09
Gitea support?


LunaTheFoxgirl:
I see there's a PR for Gogs support, have you considered Gitea support? Gogs is very slow at updating and some people has forked gogs and made gitea, which is a community effort. @s-ludwig.
Its API is way more fully fletched and should be easier to integrate.

bausshf:
I would like to see this too!

dhasenan:
I took a look at it.

[Here's Gitea's API.](https://try.gitea.io/api/swagger) Every endpoint requires authentication, which requires a user. But the API doesn't let you register a user.

The process of registering your project hosted on gitea would be:

1. Set DISABLE_REGISTRATION in your gitea config to false. Double-check that API access is turned on. Restart gitea.
2. Create a new user.
3. Enable API access for that user and disable everything else.
4. Create an API token for them.
1. Input your project URL in the dub registry.
3. Tell dub that it's hosted on gitea.
6. Give it the auth token you just made.
7. Tell it where your README is located.

There are two alternatives:

1. Dub parses gitea's HTML frontend. This will be a bit slow, and more importantly it will break without notice. Worse, this is a self-hosted system, so dub would have to support many versions of gitea HTML.
2. Dub uses git instead, costing the registry a lot more disk space, bandwidth, and time, but supporting every git repository (such as git.gnome.org).

Member1221:
You dont need DISABLE_REGISTRATION if you have admin access on the server, then you can just create a new user via the admin UI.

Member1221:
Aswell my testing shows that I don't need authentication for the GET requests. Maybe the repos you are trying to read are private?

Public repositories work fine without authentication.

dhasenan:
Apparently older versions of gitea required authentication for everything. (I'm on 1.1.3, and it's not trivial to upgrade.) https://git.ikeran.org/api/v1/repos/dhasenan/webkit-d for instance says 401 Unauthorized.

Member1221:
Well you don't need to list files afaik, just download a recent archive.

dhasenan:
Right. It's a *bit* of a problem, not a show-stopper. It means that dub needs to use more bandwidth (for itself and for the remote gitea server), time, and disk space to get the data it needs.

ljmf00:
If there is no effort to support this, I would like to try it, in a near future :smiley:

s-ludwig:
Opened a PR: #553

PR #553, opened 2023-09-24
Add Gitea support

s-ludwig:
Also Refactors the repository factory logic to enable proper matching of repository URLs to providers, so that local provider instances can be supported properly.

The `GiteaRepository` implementation is an adjusted version of `GithubRepository`, as the APIs are very similar.

### DIFF
--- a/source/dubregistry/repositories/gitea.d
+++ b/source/dubregistry/repositories/gitea.d
@@ -41,7 +41,7 @@ class GiteaRepositoryProvider : RepositoryProvider {
 		import std.conv : to;
 
 		string host = url.host;
-		if (!url.startsWith(URL(m_url)) && !host.endsWith(".gitea.com") && host != "gitea.com" && host != "gitea")
+		if (!url.startsWith(URL(m_url)))
 			return false;
 
 		repo.kind = "gitea";
@@ -70,12 +70,12 @@ class GiteaRepositoryProvider : RepositoryProvider {
 		DbRepository r;
 		assert(h.parseRepositoryURL(URL("https://example.org/foo/bar"), r));
 		assert(r == DbRepository("gitea", "foo", "bar"));
-		assert(h.parseRepositoryURL(URL("https://gitea.com/foo/bar"), r));
+		assert(h.parseRepositoryURL(URL("https://example.org/foo/bar"), r));
 		assert(r == DbRepository("gitea", "foo", "bar"));
 		assert(!h.parseRepositoryURL(URL("http://bitbucket.org/bar/baz/"), r));
-		assertThrown(h.parseRepositoryURL(URL("http://gitea.com/foo/"), r));
-		assertThrown(h.parseRepositoryURL(URL("http://gitea.com/"), r));
-		assertThrown(h.parseRepositoryURL(URL("http://gitea.com/foo/bar/baz"), r));
+		assertThrown(h.parseRepositoryURL(URL("https://example.org/foo/"), r));
+		assertThrown(h.parseRepositoryURL(URL("https://example.org/"), r));
+		assertThrown(h.parseRepositoryURL(URL("https://example.org/foo/bar/baz"), r));
 	}
 
 	Repository getRepository(DbRepository repo)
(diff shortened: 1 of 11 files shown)


source/dubregistry/repositories/gitea.d
```diff
@@ -0,0 +1,284 @@
...
+	private {
+		string m_token;
+		string m_url;
+	}
+@safe:
+
+	private this(string token, string url)
+	{
+		m_token = token;
+		m_url = url.length ? url : "https://gitea.com/";
```
WebFreak001, line 29: a lot of gitea instances are self-hosted, could we somehow make it so that this URL can be set per-package?

WebFreak001: actually nvm, since you need authentication I don't think that's gonna solve this here. Does this just exist for custom/private DUB registry instances?

s-ludwig: > Does this just exist for custom/private DUB registry instances?

Exactly. In fact, I haven't really tested this with the official gitea.com, but with a local instance.

source/dubregistry/repositories/gitea.d
```diff
@@ -0,0 +1,284 @@
...
+		addRepositoryProvider("gitea", h);
+	}
+
+	bool parseRepositoryURL(URL url, out DbRepository repo)
+	@safe {
+		import std.algorithm.searching : endsWith;
+		import std.conv : to;
+
+		string host = url.host;
+		if (!url.startsWith(URL(m_url)) && !host.endsWith(".gitea.com") && host != "gitea.com" && host != "gitea")
```
WebFreak001, line 44: since this is only for private registries, I think we probably want to not have the gitea.com / gitea shortcuts here, or at most have only the shortcut here

WebFreak001 (changes requested):
LGTM otherwise

s-ludwig:
> can you mark this PR as not fixing linked issue https://github.com/dlang/dub-registry/issues/344, since it doesn't actually cover the use-cases presented there?

I don't really understand this. Which use cases do you mean?

source/dubregistry/repositories/gitea.d
s-ludwig: It's not only for private registries, although we can probably remove this anyway. I just took it over from the original URL parsing code.

WebFreak001:
I saw in the comments people were talking about their self-hosted gitea instances

s-ludwig:
Okay, I may be wrong here, but what I assume is (and what I also do myself) that they run their own registry locally and let that point to their own Gitea instance. Everything else seems a little bit far off at the moment, and really doesn't make too much sense overall for public open-source code.

WebFreak001 (approved):
LGTM for private registries, I think this wouldn't make much sense to be configured on the public dub registry though, since gitea.com would pretty much be the only choice to configure this for and I personally would say it'd be better to in the future work on a generic GIT + dub's own storage solution, so it's not tied to any providers or web APIs at all.

s-ludwig:
> better to in the future work on a generic GIT + dub's own storage solution, so it's not tied to any providers or web APIs at all.

I'm not sure about that. On one hand, it's of course nice to have a generic approach, but on the other hand there are critical features that are only available via the API. Push notifications for changes are maybe the most critical, but having to store/cache zip archives for all projects/versions/branches could easily become a problem if there are enough projects with large binary files in their repositories.

Also, I think this would require to actually clone each repository, as opposed to just query tags/branches/files individually, which is a rather expensive operation that would either have to be repeated for each update cycle, or would have to be stored permanently. Not sure how this would look today, but, generally speaking, this is surely not a viable solution in terms of hosting costs.

s-ludwig:
Having said that, maybe if this is just an additional option, where we automatically recognize github/gitlab/bitbucket git URLs and actually use the API for those, and only use the actual git protocol for those where that is not possible, it might be more realistic. On the other hand, there is a lot of merit in forcing projects to be in a more public place in terms of enabling a more healthy community interaction (bug reports, pull requests).

PR #573, opened 2024-06-12
Remove dpldocs link, it no longer functions.

schveiguy:
image

DPLDocs no longer functions as a reference. This removes the link, it can be re-added if it starts working again in the future.

### DIFF
--- a/views/view_package.dt
+++ b/views/view_package.dt
@@ -111,7 +111,6 @@ block body
 			- auto doc_url = packinfo.info["documentationURL"].get!string;
 			- if (doc_url.length)
 				a.tab.external(href=doc_url, target="_blank") Documentation
-			a.tab.external(href=text("https://",normalizedPackageName,".dpldocs.info/",versionInfo["version"].get!string,"/"), target="_blank") Reference
 		- bool renderedTabContent;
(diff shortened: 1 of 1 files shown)


Geod24:
Just remove the link ? We can git revert if it opens back up.

schveiguy:
OK, done

WebFreak001:
the tool is very useful, maybe we can just finance it through the DLF?

schveiguy:
I'm not sure what the DLF wants to do, but I'd suggest this isn't worth the cost. We are better off having people use github to host their docs (they can even use adrdox to do it, e.g. https://pbackus.github.io/sumtype/sumtype.html)

In any case, I'm just fixing a broken link. If the DLF decides to finance it, we can put it back. We should not have a broken link that you *can't turn off* on every page.

WebFreak001:
forced auto-generated documentation for packages is worth the very tiny server cost imo - this is good for SEO and also just encourages people to write more documentation because it's low effort for them and just works out of the box

schveiguy:
I think it's fine to have it if the link works. But having a broken link isn't good, even for SEO.

We can merge this now, and then merge the revert later if the site opens up.

PR #587, opened 2025-03-22
Add library intro

Khushi281300:
This PR includes several updates and improvements to various parts of the project. Key changes include:

Updated dependencies in dub.selections.json (e.g., diet-ng updated to version 1.8.3).

Fixed a broken link in source/dubregistry/viewutils.d (corrected link to package scoring documentation).

Refactored the front-end code in views/home.dt to improve the layout and structure, including category selection enhancements.

Updated navigation and package display logic across multiple views (views/layout.inc.menu.dt, views/view_package.dt).

### DIFF
--- a/views/home.dt
+++ b/views/home.dt
@@ -1,122 +1,135 @@
 extends layout
 
 block title
-	- import dubregistry.viewutils;
-	- import dubregistry.web;
-	- import vibe.data.json;
-	- import std.algorithm.comparison : min;
-	- import std.algorithm.iteration : map;
-	- import std.string : replace, split;
-	- title = "Find, Use and Share DUB Packages";
-	script(type="application/javascript", src="scripts/home.js")
-	script(type="application/javascript").
-		window.categories = #{serializeToJson(info.categories).toString()};
+    - import dubregistry.viewutils;
+    - import dubregistry.web;
+    - import vibe.data.json;
+    - import std.algorithm.comparison : min;
+    - import std.algorithm.iteration : map;
+    - import std.string : replace, split;
+    - title = "Find, Use and Share DUB Packages";
+    script(type="application/javascript", src="scripts/home.js")
+    script(type="application/javascript").
+        window.categories = #{serializeToJson(info.categories).toString()};
 
 block body
-	- auto category = req.query.get("category", null);
-	- auto sort_mode = req.query.get("sort", "updated");
+    - auto category = req.query.get("category", null);
+    - auto sort_mode = req.query.get("sort", "updated");
 
-	- string makeHomeURL(string sort, string category, ulong skip, ulong limit) {
-	-   import std.conv : to;
-	-   return "?sort=" ~ sort ~ "&category=" ~ category ~ "&skip=" ~ skip.to!string ~ "&limit=" ~ limit.to!string;
-	- }
+    - string makeHomeURL(string sort, string category, ulong skip, ulong limit) {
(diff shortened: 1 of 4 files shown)


WebFreak001 (changes requested):
please don't mix changes with indentation/formatting changes such as the tabs -> spaces change that you did, this makes it hard to review

There is a home-fixed.dt file which you probably didn't intend to include

The change to viewutils.d was done in a separate PR and is merged already, please rebase

khushishukla2813:
> please don't mix changes with indentation/formatting changes such as the tabs -> spaces change that you did, this makes it hard to review
> 
> There is a home-fixed.dt file which you probably didn't intend to include
> 
> The change to viewutils.d was done in a separate PR and is merged already, please rebase

i will rebase,for indentaion i tried to make it like before , but it's showing error.
image


should i remove indentation? when i am fixing it as above then no error..

WebFreak001:
the error shows you are mixing tabs and spaces, please use them consistently. Currently the files in this repo use tabs so it'd be better to convert your changes all to tabs

views/home.dt
```diff
@@ -27,6 +27,18 @@ block body
 	- else
 		p Welcome to DUB, the D package registry. The following list shows all available packages:
 
+	//New Section
```
WebFreak001, line 30: please don't add comments that refer to old/new things, this comment will be outdated immediately on the next change and isn't helping anyone. I would just suggest removing this line since the `h2` describes it already

views/layout.inc.menu.dt
```diff
@@ -21,8 +21,8 @@
 			ul
 				- foreach(items; listitems)
 					- if(items[1].length < 2)
-						- bool active = startsWith(req.path[1..$], items[1][0]);
-						- if( items[1][0].empty ) active = req.path == "/";
+						- bool active = startsWith(req.requestPath.toString()[1..$], items[1][0]);
```
WebFreak001, line 24: this is unrelated to the change above, please only do one thing at a time in a PR, you can have multiple PRs open at the same time

views/home.dt
```diff
@@ -27,6 +27,18 @@ block body
...
+	h2 📚 Commonly Used Libraries
+	ul
+		li
+			a(href="https://vibed.org/") 🌐 vibe.d - A high-performance asynchronous I/O web framework
+		li
+			a(href="http://gtkd.org/") 🖥️ gtkd - GTK bindings for building GUI applications in D
+		li
+			a(href="https://github.com/BindBC") 🔗 bindbc - A collection of bindings for C libraries
+		li
+			a(href="https://github.com/dlang-community/gettext") 🌎 gettext - A library for handling internationalization
```
WebFreak001, line 40: I personally like having a list of common libraries, but I don't know if the registry is the right place for this. This needs some discussion on the forum where to put it and which libraries to include in it I think. There are already a lot of "awesome" lists for D projects, e.g. https://github.com/dlang-community/awesome-d/ - it should probably rather link to a project like this instead of maintaining a separate list.

We already have the dynamically updating "Most popular" listing as well. It would probably make more sense to update the scoring there if you think certain packages shouldn't be ranked so high.

Also I don't think it's a good idea to use emoji on websites due to possible encoding issues, not all fonts having them and the look being different on each platform / emoji font which isn't really controllable and quickly gets ugly. For categories we already included monochrome icons which could be reused here instead.

WebFreak001 (changes requested):

PR #598, opened 2025-12-26
Add subpackages to the index

dejlek:
This will in turn let users search through subpackages as well, so when someone searches for `gtk4` it should give `gid` as it has `gid:gtk4` subpackage.

### DIFF
--- a/source/dubregistry/dbcontroller.d
+++ b/source/dubregistry/dbcontroller.d
@@ -57,22 +57,22 @@ class DbController {
 		m_downloads.createIndex(IndexModel().add("package", 1).add("version", 1));
 
 		// add current text index
- 	immutable keyWeights = [
- 		"name": 8,
- 		"categories": 4,
- 		"versions.info.subPackages.name": 4,
- 		"versions.info.description": 2,
- 		"versions.info.authors": 1
- 	];
- 	Bson[string] fts;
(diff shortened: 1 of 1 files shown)


source/dubregistry/dbcontroller.d
```diff
@@ -369,6 +370,24 @@ class DbController {
 		foreach (ref pkg; pkgs)
 			pkg.textScore = (pkg.textScore - minMaxTS[0]) * scale + DbPackageStats.minScore;
 
+		// Also search for substring matches in package name using regex
+		auto regexPkgs = m_packages
+			.find(["name": ["$regex": query, "$options": "i"]])
```
WebFreak001: missing regex escape + max length, this could otherwise lead to a DoS

You added the search to the full-text-search, is that not sufficient for the results?

WebFreak001 (changes requested):
otherwise LGTM

source/dubregistry/dbcontroller.d
dejlek: Thanks for the feedback. Element took my PR fixed the regex issue you mentioned and added some more, so I will close this.

WebFreak001: um I don't think there is a PR opened for this yet though?

PR #599, opened 2025-12-27
Add ForgejoRepositoryProvider and corresponding config options

s-ludwig:
This allows to configure codeberg.org as a repository base URL for the registry.


dejlek:
I thought it will work with gitea. Thanks for looking into this, @s-ludwig !

s-ludwig:
@dejlek: The problem is just that currently there is a one-to-one relationship between repository provider type and site URL, so that only one of gitea.com and codeberg.com would work that way.

WebFreak001 (approved):
more flexible selection could be added in the future, but first having the provider seems good

WebFreak001:
btw, off-topic, are you at 39C3 by any chance? - contact me in slack or call me on DECT if you are ^^

WebFreak001:
frontend still seems to need some tweaks (external link to repo, icon for codeberg/forgejo)

s-ludwig:
@WebFreak001: I've fixed the links in #601, the icon looks like it's going to be a bit more involved (update of the fontello font with a custom SVG icon).

PR #600, opened 2025-12-28
Improved search with sub-string and sub-package support

elementgreen:
Improves the search feature by searching sub-package names and doing sub-string queries on package and sub-package names.  Search results will show sub-packages.  After various optimizations, the search feature is only about 3X slower even with the additional escaped regex queries.  Benchmarked with random 4 character or less query strings.

- Optimized original search by creating a smaller DbSearchPackage structure and limiting queries to latest version
- Add an index on versions.info.subPackages.name
- Add versions.info.subPackages.name to FTS index with a weight of 6, so that full string matches of sub-packages gets a higher textScore than sub-string matches
- Update FTS index to packages_full_text_search_index_v4
- Package name sub-string match gets a score of 7, placing it just below a full string match (8)
- Sub-package name sub-string match gets a score of 5, just under a full string match (6)

### DIFF
--- a/source/dubregistry/dbcontroller.d
+++ b/source/dubregistry/dbcontroller.d
@@ -349,6 +349,11 @@ class DbController {
 
 	DbSearchPackage[] searchPackages(string query)
 	{
+		query = query.strip;
+
+		if (query.length > 255) // Limit length to prevent DoS
+			query = query[0 .. 256];
+
 		auto project = [
 			"name": Bson(1),
 			"stats": Bson(1),
@@ -359,7 +364,7 @@ class DbController {
  		auto hasVersionFilter = Bson(["$match": Bson(["versions": Bson(["$exists": Bson(true),
 			"$ne": Bson(cast(Bson[])[])])])]);
 
-		if (!query.strip.length) {
+		if (!query.length) {
 			return m_packages
 				.aggregate(
 					hasVersionFilter,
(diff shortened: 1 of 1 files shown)


elementgreen:
Just for the record.. I think 3X slower is probably acceptable for the additional sub-string and sub-package matching.  If additional performance was desirable I'd probably look into constructing a search table that is synchronized to the packages table and contains package/full sub-package names and the minimal required set of data for the latest version. This could be searched in one pass using regex, without having to unwind the subpackages and concat the package names.

Other things I tried to see if they could improve performance:
- Using the $facet command to parallel the FTS and regex searches. Didn't work because FTS textScore metadata not retrievable for the $project output and the $match for FTS must come first in the aggregate which results in it also being applied to the other searches, essentially disabling the sub-string searches.
- Using the $facet command to run the two regex searches in parallel.  Ended up being 6X slower!  Seems there are known performance issues with $facet in MongoDB.
- Looked into potentially paralleling the 2 regex sub-string searches in D, but it wouldn't save much because the sub-package search is the bulk of the time compared to the package sub-string search (25ms vs 3ms on my system).
- Trimmed down the returned data further, by filtering out unused fields in the latest version.  The minor increase in performance (maybe 1% or 2%) seemed not worth it compared to the additional $project complexity added.

s-ludwig (approved):
Looks good AFAICS. The performance degradation is a bit unfortunate, since it's already kind of slow, but if everything otherwise works as intended, it definitely seems worth the trade-off.

dejlek:
@s-ludwig who can merge/deploy this?

PR #603, opened 2025-12-30
Add webhooks for package updating

### LINKED ISSUE #412: Update package webhook

WebFreak001:
Fixes #412, supersedes #414


thewilsonator (approved):

CyberShadow:
Thank you !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!


This seems to be a broken link right now: https://github.com/dlang/dub/blob/master/docs/http-update.md

image

CyberShadow:
It would be nice if there were readily available instructions for how to set this up correctly - i.e. what exactly to fill out in this form:

image

CyberShadow:
More issues:

- The "GitHub Webhook" URL is invalid, it looks like `https://code.dlang.org:443?secret=xxxxxxxxxxxxxxxxxxxxxx/api/packages/xxxxx/update/github?secret=xxxxxxxxxxxxxxxxxxxxxx`

  Note how `?secret=xxxxxxxxxxxxxxxxxxxxxx` appears twice, corrupting the URL.
   
- It should probably mention somewhere that you *must* set the content type to JSON, otherwise it will just return a 400 error.

PR #616, opened 2026-07-18
Render AsciiDoc package READMEs on the Info tab

AMDphreak:
## Summary
- Prefer `README.md`, then `README.adoc` / `README.asciidoc`, then bare `README`, then other `README*` files.
- Store `readmeFormat` / `readmeFile` and render AsciiDoc via [asciidoctor-d](https://github.com/dlang-supplemental/asciidoctor-d) (secure HTML fragment).
- Sanitize README HTML (strip script/style/handlers) and add light AsciiDoc CSS.
- Preserve historical default of Markdown filtering for older DB documents without `readmeFormat`.

Depends on fragment/secure support in asciidoctor-d (`feature/html5-fragment-safe`). Until that lands on a tagged release, `dub.sdl` points at that branch.

## Test plan
- [ ] Package with only `README.adoc` shows formatted Info tab (headings, lists, code)
- [ ] Package with `README.md` unchanged
- [ ] AsciiDoc passthrough / `javascript:` links are neutralized
- [ ] Relative image/link rewriting still works for GitHub packages

Made with [Cursor](https://cursor.com)

### DIFF
--- a/dub.sdl
+++ b/dub.sdl
@@ -11,8 +11,10 @@ dependency "botan" version="*" # needs bcrypt, available since initial release
 dependency "dub" version="~>1.33"
 dependency "userman" version="~>0.4.0"
 dependency "uritemplate" version="~>1.0.0"
+dependency "asciidoctor-d" version="~feature/html5-fragment-safe" repository="git+https://github.com/dlang-supplemental/asciidoctor-d.git"
 subConfiguration "dub" "library-nonet"
 subConfiguration "botan" "full" # full = everything excluding openssl (needed for bcrypt)
+subConfiguration "asciidoctor-d" "library"
 
 //versions "VibeJsonFieldNames"
 
--- a/public/styles/asciidoc.css
+++ b/public/styles/asciidoc.css
@@ -1,0 +1,83 @@
+/* Lightweight AsciiDoc fragment styles for the package Info tab */
+
+.asciidoc-body #header h1 {
+  margin-top: 0;
+}
+
+.asciidoc-body .sect1 + .sect1,
+.asciidoc-body .sect2 + .sect2 {
+  margin-top: 1.5em;
+}
+
+.asciidoc-body .listingblock,
+.asciidoc-body .literalblock {
+  margin: 1em 0;
+}
+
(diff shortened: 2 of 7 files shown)


dub.sdl
```diff
@@ -11,8 +11,10 @@ dependency "botan" version="*" # needs bcrypt, available since initial release
 dependency "dub" version="~>1.33"
 dependency "userman" version="~>0.4.0"
 dependency "uritemplate" version="~>1.0.0"
+dependency "asciidoctor-d" version="~feature/html5-fragment-safe" repository="git+https://github.com/dlang-supplemental/asciidoctor-d.git"
```
0xEAB, line 14: This dependency should probably get registered at code.dlang.org.
More importantly, DUB’s selections file has not been updated to reflect this change.

source/dubregistry/viewutils.d
```diff
@@ -129,3 +130,101 @@ unittest {
...
+}
+
+/** Strip dangerous tags/attributes from README HTML (defense in depth). */
+string sanitizeReadmeHtml(string html)
+{
+	import std.regex : ctRegex, replaceAll;
+
+	// Drop script/style/iframe/object/embed/link/meta/form controls and their contents.
+	static dangerous = ctRegex!(
+		`(?is)<\s*(script|style|iframe|object|embed|link|meta|form|input|button|textarea|select)(\s[^>]*)?>.*?<\s*/\s*\1\s*>|<\s*(script|style|iframe|object|embed|link|meta|form|input|button|textarea|select)(\s[^>]*)?/?>`);
```
0xEAB, line 213: I’m not keen on such a block-list approach.

Also, wasn’t it general wisdom to never use regex for parsing XML or HTML? Has this changed?

public/styles/asciidoc.css
```diff
@@ -0,0 +1,83 @@
+/* Lightweight AsciiDoc fragment styles for the package Info tab */
```
0xEAB, line 1: Why do they need their own distinct styling? Should they rather be styled in tandem with markdown documents?

0xEAB (changes requested):

PR #618, opened 2026-08-08
Add GET /my_packages/:name/webhook status endpoint

### LINKED ISSUE #617: Add non-destructive webhook status endpoint for package admins
## Problem

Package admins (and owner CLIs such as [dub-publish](https://github.com/dlang-supplemental/dub-publish)) need to know whether a webhook secret is already configured **without regenerating it**.

Today the only owner HTTP actions around webhooks are:

- `POST /my_packages/:name/regen_secret` — **creates or rotates** the secret (destructive; forge webhook URLs that embed `?secret=` go stale)
- `POST /my_packages/:name/unset_secret` — disables webhooks

The My packages HTML page can show that a secret exists (Disable / Regenerate buttons), but there is no stable machine-readable status endpoint. Tooling either scrapes HTML or accidentally calls `regen_secret` when checking status.

Plaintext secrets correctly remain one-shot (shown only right after regenerate). This request is only for a boolean `configured` flag.

## Proposal

Add an authenticated, non-destructive endpoint for package admins:

`GET /my_packages/:packageName/webhook`

Response:

```json
{package:mypkg,configured:true}
```

Same session auth as other `/my_packages/...` owner routes. Never return the plaintext secret.

## Context

Working around this in dub-publish led to a UX bug: bare `hooks` always regenerated the secret. Status vs generate is now split client-side, but a first-class registry API would remove the HTML scrape.

AMDphreak:
## Summary
- Add authenticated `GET /my_packages/:packname/webhook` returning JSON `{package,configured}` without regenerating the secret
- Add `hasPackageSecret` on the DB/registry layers
- Document the endpoint in `api-docs/http-update.md`

Fixes https://github.com/dlang/dub-registry/issues/617

## Motivation
Owner tooling (e.g. dub-publish) needs to check whether webhooks are enabled. The only existing owner HTTP actions are `regen_secret` (destructive rotate) and `unset_secret`. Scraping the My packages HTML works but is fragile; calling `regen_secret` to check status invalidates forge webhook URLs.

Plaintext secrets remain one-shot after regenerate — this endpoint only exposes a boolean.

## Test plan
- [ ] As package admin, `GET /my_packages/:name/webhook` with session cookie returns `configured:false` before enable
- [ ] After Enable Webhooks / `regen_secret`, same GET returns `configured:true`
- [ ] Non-admin / unauthenticated requests are rejected
- [ ] Response never includes the plaintext secret

### DIFF
--- a/api-docs/http-update.md
+++ b/api-docs/http-update.md
@@ -2,6 +2,20 @@
 
 To queue an update of your package you can use the `POST /api/packages/:packageName/update` endpoint.
 
+## `GET /my_packages/:packageName/webhook` (authenticated)
+
+Returns whether a webhook secret is configured for the package **without regenerating it**.
+
+Requires a logged-in package admin session (same auth as the My packages UI / `regen_secret`).
+
+Response JSON:
+
+```json
+{"package":"mypkg","configured":true}
+```
+
+The plaintext secret is intentionally never returned here — it is only shown once after `POST .../regen_secret`.
+
 ## `POST /api/packages/:packageName/update`
 
 Queues an update for the specified package.
--- a/source/dubregistry/dbcontroller.d
+++ b/source/dubregistry/dbcontroller.d
@@ -227,6 +227,15 @@ class DbController {
 		else return validateBcryptHash(ret.get.get!string, secret);
 	}
 
+	/// True when a webhook secret hash is stored (does not reveal the plaintext).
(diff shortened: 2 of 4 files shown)


thewilsonator:
cc @Geod24 this looks fine to me, but it would be good to get a second pair of eyes given this touches `@auth`

